PR #59, opened 2022-09-13
feat: add support for ETSRSS marker

### LINKED ISSUE #51: [MARKER]: ETS_RSS

edge-wise time-series RSS values

Inputs:
4D nifti, processed (e.g. GS regression)

Output:
An array of length of number of time-points

https://www.pnas.org/doi/10.1073/pnas.2005531117

```shell
@leo
```

### Anything else to say?

_No response_

LeSasse:
* [x] fix #51
* [x] tests added/passed

We have added the RSSETSMarker to compute the root sum of squares of the edge-wise time series derived from the z-scored BOLD time series based on the paper by [Zamani Esfalahni et al. (2020)](https://doi.org/10.1073/pnas.2005531117).


synchon:
Need to implement `storage.store_timeseries()` before getting coverage up.

docs/changes/latest.inc
```diff
@@ -10,7 +10,7 @@
 
 .. NOTE: add the contributors and reference to the github issue/PR at the end
    Example:
-      - Implemented feature X (:gh:`151` by `Sami Hamdan`_).
+   - Implemented feature X (:gh:`151` by `Sami Hamdan`_).
```
fraimondo, line 13: Is this still part of the note? Or it will become a new element in the list?

junifer/markers/ets_rss.py
```diff
@@ -0,0 +1,124 @@
...
+        list of str
+            The updated list of output kinds, as storage possibilities.
+
+        """
+        outputs = []
+        for t_input in input:
+            if t_input in ["BOLD"]:
+                outputs.append("timeseries")
+            else:
+                raise ValueError(f"Unknown input kind for {t_input}")
```
fraimondo, line 63: This is not needed here.

fraimondo: As it is only one input supported, it should only return ["timeseries"]

junifer/markers/ets_rss.py
```diff
@@ -0,0 +1,124 @@
...
+        Parameters
+        ----------
+        kind
+        out
+        storage
+
+        """
+        logger.debug(f"Storing {kind} in {storage}")
+        if kind in ["BOLD"]:
+            storage.store_timeseries(**out)
```
fraimondo, line 79: Include this in coverage, otherwise we don't know if it works.

fraimondo (changes requested):

fraimondo:
Codecov not happy with this PR

docs/changes/latest.inc
synchon: I wasn't sure about this either and just fixed the indentation.

junifer/markers/ets_rss.py
synchon: Makes sense.

junifer/markers/ets_rss.py
synchon: The `store_timeseries()` method fails as it's not implemented, or did you implement it?

docs/changes/latest.inc
LeSasse: I thought fede added this as an example?

junifer/markers/ets_rss.py
LeSasse: I do want to make a small change so that the edge time series is computed in a little helper function _ets() that takes an ndarray and returns an ndarray, because I also need this for the EdgeCentricFC marker, which I will then put in the same file, which I want to rename ets.py perhaps.

LeSasse:
I do want to make a small change so that the edge time series is computed in a little helper function _ets() that takes an ndarray and returns an ndarray, because I also need this for the EdgeCentricFC marker, which I will then put in the same file, which I want to rename ets.py perhaps.

synchon:
I would make a new module `junifer/markers/utils.py` and add the helper function there.

LeSasse:
> I would make a new module `junifer/markers/utils.py` and add the helper function there.

I have added this. One problem that I had when figuring out coverage is that when I appended `-missing` to that line in tox.ini and ran ```tox -e coverage``` the result told me that lines 16 - 64 were missing, but those are included in the tests, so I am not sure what is going on there.

junifer/markers/ets_rss.py
LeSasse: I have removed this.

LeSasse:
Also, I am not sure why the CI is complaining about linting. When I run flake8 it seems all fine. Is there something wrong in my settings?

synchon:
> > I would make a new module `junifer/markers/utils.py` and add the helper function there.
> 
> I have added this. One problem that I had when figuring out coverage is that when I appended `-missing` to that line in tox.ini and ran `tox -e coverage` the result told me that lines 16 - 64 were missing, but those are included in the tests, so I am not sure what is going on there.

That's weird, let me check it on my system.

synchon:
> Also, I am not sure why the CI is complaining about linting. When I run flake8 it seems all fine. Is there something wrong in my settings?

Those are from the new files.

synchon:
@LeSasse let me know when you are done with your work, I can do the necessary rebasing and get it to pass the CI.

synchon:
> Ah yes, so subjectively i am happy with this code now, and do not plan any changes on this unless there are some requests

Awesome! Let me rebase and check the CI.

docs/changes/latest.inc
synchon: Reverting the change.

fraimondo:
@synchon: the file `test_utils.py` needs to be renamed to `test_marker_utils.py`. Otherwise pytest gets confused as there is another `test_utils.py` file (for `junifer.utils`)

synchon:
Yeah doing that.

synchon:
@fraimondo I pushed an extra commit for linting, ignore that commit for review.

fraimondo (approved):

PR #93, opened 2022-09-30
update: add DataladHCP1200 datagrabber test

### LINKED ISSUE #4: [ENH] HCP Test Data Set on Gin
We need to create a test data set on Gin (https://gin.g-node.org/G-Node/Info/wiki) that is modeled after the real HCP open access datalad dataset (https://github.com/datalad-datasets/human-connectome-project-openaccess), so that we can test our HCP datalad datagrabbers automatically using github.

synchon:
* [x] fix #4
* [x] tests added/passed

This PR adds test for DataladHCP1200 datagrabber and thus also adds script for test dataset generation on Gin.

### DIFF
--- a/junifer/datagrabber/hcp.py
+++ b/junifer/datagrabber/hcp.py
@@ -174,8 +174,6 @@ class DataladHCP1200(DataladDataGrabber, HCP1200):
     phase_encodings : {"LR", "RL"} or list of the options, optional
         HCP phase encoding directions. If None, both will be used
         (default None).
-    check_file_existence : bool, optional
-        Whether to check if file exists while accessing (default False).
     **kwargs
         Keyword arguments passed to superclass.
 
@@ -186,7 +184,6 @@ class DataladHCP1200(DataladDataGrabber, HCP1200):
         datadir: Union[str, Path, None] = None,
         tasks: Union[str, List[str], None] = None,
         phase_encodings: Union[str, List[str], None] = None,
-        check_file_existence: bool = False,
         **kwargs,
     ) -> None:
         """Initialize the class."""
@@ -199,7 +196,6 @@ class DataladHCP1200(DataladDataGrabber, HCP1200):
             datadir=datadir,
             tasks=tasks,
             phase_encodings=phase_encodings,
-            check_file_existence=check_file_existence,
             uri=uri,
             rootdir=rootdir,
         )
--- a/junifer/datagrabber/hcp.py
+++ b/junifer/datagrabber/hcp.py
@@ -174,6 +174,8 @@ class DataladHCP1200(DataladDataGrabber, HCP1200):
     phase_encodings : {"LR", "RL"} or list of the options, optional
         HCP phase encoding directions. If None, both will be used
         (default None).
+    check_file_existence : bool, optional
+        Whether to check if file exists while accessing (default False).
     **kwargs
         Keyword arguments passed to superclass.
 
@@ -184,6 +186,7 @@ class DataladHCP1200(DataladDataGrabber, HCP1200):
         datadir: Union[str, Path, None] = None,
         tasks: Union[str, List[str], None] = None,
         phase_encodings: Union[str, List[str], None] = None,
+        check_file_existence: bool = False,
         **kwargs,
     ) -> None:
         """Initialize the class."""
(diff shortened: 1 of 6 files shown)


fraimondo:
It needs a bit more of testing.

The HCP has the possibility of getting several tasks and phase encodings. These are also parameters in the constructor, so to restrict the iterator to certain tasks or phase_encodings. So far the test covers only the first element. @synchon: can you test setting the task / phase encoding parameters and that the iterator works accordingly?

synchon:
Ok sure taking a look.

fraimondo:
So the parametrization test is good to test that all the tasks/phase encoding pairs are well formed. The problem is the iterator. If you set phase/tasks in the constructor of the datagrabber, the iterator should restrict to these elements only. Can you make a test for it?

synchon:
Sure.

fraimondo:
testing the bot

synchon:
> So the parametrization test is good to test that all the tasks/phase encoding pairs are well formed. The problem is the iterator. If you set phase/tasks in the constructor of the datagrabber, the iterator should restrict to these elements only. Can you make a test for it?

I think I have addressed the iterator restriction test. Let me know if it is okay.

junifer/datagrabber/hcp.py
```diff
@@ -196,6 +199,7 @@ def __init__(
             datadir=datadir,
             tasks=tasks,
             phase_encodings=phase_encodings,
+            check_file_existence=check_file_existence,
```
fraimondo, line 202: This does not need to be here. Only in the pattern one.

junifer/datagrabber/tests/test_hcp.py
```diff
@@ -0,0 +1,142 @@
...
+        ("WM", "LR"),
+        ("WM", "RL"),
+        ("RELATIONAL", "LR"),
+        ("RELATIONAL", "RL"),
+        ("EMOTION", "LR"),
+        ("EMOTION", "RL"),
+        ("LANGUAGE", "LR"),
+        ("LANGUAGE", "RL"),
+        ("GAMBLING", "LR"),
+        ("GAMBLING", "RL"),
```
fraimondo, line 108: Can you try a couple of tasks too? Also one task, two phase encodings?

```
(["REST1", "REST2"], ["LR", "RL"]),
("REST1", ["LR", "RL"]),
(["REST1", "REST2"], None),
```

fraimondo (changes requested):

junifer/datagrabber/hcp.py
synchon: Isn't it safe to keep it here as well considering it's the HCP datagrabber?

fraimondo:
Just two small changes, and test for errors if tasks/phase encodings are wrong.

junifer/datagrabber/tests/test_hcp.py
synchon: The cases you commented should be good for the test right? I mean, we just need a few cases.

fraimondo: exactly

synchon: Done.

junifer/datagrabber/hcp.py
synchon: Or maybe we explicitly pass `False` here?

synchon: Okay on second thought, I will remove it, doesn't really harm.

synchon: Done.

junifer/datagrabber/hcp.py
```diff
@@ -199,3 +199,8 @@ def __init__(
             uri=uri,
             rootdir=rootdir,
         )
+
+    @property
```
fraimondo, line 203: Do we need to have this here? I thought the pattern one would have the default (`False`)

fraimondo: Ah, sorry, this is the Datalad one, this should return `True` then.

fraimondo (changes requested):

junifer/datagrabber/hcp.py
synchon: I believe this is ok then.

Issue #74, opened 2022-09-14
[DOC]: Fix and complete docstrings + types


fraimondo:
### Which element of the documentation do you want to modify and why?

The hackathon has left us with a bunch of methods in which the typing and docstrings might be incomplete.

PR #103, opened 2022-10-14
fix: storage

### LINKED ISSUE #84: [BUG]: Fix storage functions
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We have `store_timeseries`, `store_matrix2d`, `store_df`... etc. Everything is a mess right now.

### Expected Behavior

I expect to have a simple API to store

### Steps To Reproduce

Try to create a marker that works with both VBM and BOLD, impossible to test storage.

### Environment

```markdown
Junifer main
```

### Relevant log output

_No response_

### Anything else?

_No response_

synchon:
* [x] fix #84, #74
* [x] tests added/passed

This PR improves the storage interface with a single method on `storage-like` object intuitively called `store()`. It also fixes the interaction with the markers and lets the marker choose the `kind` of storage needed. The `storage-like` object is responsible for implementing the necessary methods which `marker.store()` and finally `storage.store()` will call for example, `store_timeseries()`, `store_table()` and `store_matrix()`.
As this was a major refactor for the internals, a lot of other minor changes had to implemented including method name changes, type annotations and docstring upgrades.


junifer/markers/collection.py
```diff
@@ -104,8 +119,8 @@ def validate(self, datagrabber) -> None:
 
         for marker in self._markers:
             logger.info(f"Validating Marker: {marker.name}")
-            m_data = marker.validate(t_data)
+            m_data = marker.validate_input(t_data)
```
fraimondo, line 122: This should still be validate.

There are two methods:

`validate_input`: checks the input
`get_output_kind`: gives the output kind (given the input)

`validate` calls both of them

junifer/markers/collection.py
```diff
@@ -104,8 +119,8 @@ def validate(self, datagrabber) -> None:
...
             logger.info(f"Marker output type: {m_data}")
             if self._storage is not None:
                 logger.info(f"Validating storage for {marker.name}")
-                self._storage.validate(m_data)
+                self._storage.validate_input(m_data)
```
fraimondo, line 126: Same here, the method should be `validate`

junifer/storage/base.py
```diff
@@ -51,9 +51,8 @@ def get_meta(self) -> Dict:
         }
         return meta
 
-    # TODO: is raising ValueError required?
     @abstractmethod
-    def validate(self, input_: List[str]) -> bool:
```
fraimondo, line 56: It still should be the same: `validate`. However, the storage has no output.

junifer/storage/base.py
```diff
@@ -103,7 +97,7 @@ def list_features(
         )
 
     @abstractmethod
-    def read_df(
+    def read_features(
```
fraimondo, line 100: This is kind of tricky.

`read_features` should be able to read any kind of features: tables, matrix, nifti, etc.

The previous name `read_df`, aimed to tell you which kind of object you were getting.

I'm sill not sure about this change. We might need to discuss it in the channel.

fraimondo (changes requested):

junifer/markers/collection.py
synchon: Okay makes sense since it comes from `PipelineStepMixin`.

junifer/storage/base.py
synchon: So it raises an error if `validate` fails?

synchon: From what I see, `validate` in `SQLiteFeatureStorage` returns a bool.

junifer/storage/base.py
synchon: I see your point.

fraimondo: yes, at it really depends on the storage.

junifer/storage/base.py
fraimondo: It should not. It should raise an error if it fails to validate the action.

synchon: Okay then I'll make the changes.

junifer/markers/collection.py
synchon: Done.

synchon:
@fraimondo waiting for your approval.

fraimondo:
I'll check it now (testing the bot)

junifer/storage/base.py
```diff
@@ -140,108 +134,81 @@ def store_metadata(self, meta: Dict) -> str:
...
+        **kwargs
+            The keyword arguments.
+
+        """
+        if kind == "matrix":
+            self.store_matrix(**kwargs)
+        elif kind == "timeseries":
+            self.store_timeseries(**kwargs)
+        elif kind == "table":
+            self.store_table(**kwargs)
```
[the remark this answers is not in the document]
synchon: Done.

junifer/storage/sqlite.py
```diff
@@ -232,31 +233,35 @@ def store_2d(
...
 
         """
         # Convert input to list
         if not isinstance(input_, list):
             input_ = [input_]
 
-        return all(x in self._valid_inputs for x in input_)
+        if not all(x in self._valid_inputs for x in input_):
+            raise_error(
+                msg=(
```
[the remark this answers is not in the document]
synchon: Done.

fraimondo (approved):

PR #33, opened 2022-10-19
[ENH]: Implement a command to print the running environment details

synchon:

I've seen this in datalad and it's a very cool feature. When you run `datalad wtf`, you get a verbose print of all the characteristics of the runtime environment. Thus, whenever you submit an issue, you can copy/paste the output to help the developers understand what's going on.

As a new command in `junifer.api.cli` so we can run `junifer wtf`

```shell
No
```

### Anything else to say?

We can check datalad, but beware of the license. Maybe we can also get an approval to copy/paste and release under AGPLv3.

### DIFF
--- a/junifer/api/utils.py
+++ b/junifer/api/utils.py
@@ -74,7 +74,7 @@ def _get_dependency_information(long_: bool) -> Dict[str, str]:
             "nibabel",
             "nilearn",
             "sqlalchemy",
-            "pyyaml",
+            "yaml",
         ]:
             if key in dependency_versions.keys():
                 pruned_dependency_versions[key] = dependency_versions[key]
--- a/junifer/api/utils.py
+++ b/junifer/api/utils.py
@@ -96,8 +96,13 @@ def _get_system_information() -> Dict[str, str]:
     }
 
 
-def _get_environment_information() -> Dict[str, str]:
+def _get_environment_information(long_: bool) -> Dict[str, str]:
     """Get system environment information.
+
+    Parameters
+    ----------
+    long_ : bool
+        Whether to report long version.
 
     Returns
     -------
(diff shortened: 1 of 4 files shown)


synchon:
@fraimondo Okay so I have a basic version working with motivation from datalad. I have kept it focused on our use-case as `datalad wtf` has a lot of stuff which is better run from there. Let me know what you think. After an initial review, I'll add the tests and it should be good to go.

junifer/api/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    # Report long version
+    if long_:
+        for key, value in dependency_versions.items():
+            # Ignore built-in modules and self
+            if value != "None" and key != "junifer":
+                pruned_dependency_versions[key] = value
+
+    # Report short version
+    else:
+        for key in [
```
fraimondo, line 69: Can we read this from the pyproject.toml file?

Or we need to manually hardcode here?

I have the impression that it will end-up like the `_version.py` file, which is created on build/install

fraimondo (commented):
Can we do what I mention in my comment? Or its way too complicated for now?

We can either create a low-priority issue or code it now.

junifer/api/utils.py
synchon: The only reason I want to avoid reading from `pyproject.toml` is because of the extra I/O. The dependencies are not going to change frequently so hardcoding it doesn't seem like a bad idea.

I don't really get the analogy with the `_version.py` though.

fraimondo: On build/install time, a file `_dependencies.py` is created with an array that contains the dependencies.

Soon, we will start having "conditional" dependencies in which some specific markers will require some dependencies. I'm just trying to avoid having always in mind that we need to manage dependencies in two different sites.

fraimondo: e.g.: `pip install junifer[mri]` `pip install junifer[eeg]` `pip install junifer[surf]` `pip install junifer[dmri]` ... etc

synchon: > On build/install time, a file `_dependencies.py` is created with an array that contains the dependencies.
> 
> Soon, we will start having "conditional" dependencies in which some specific markers will require some dependencies. I'm just trying to avoid having always in mind that we need to manage dependencies in two different sites.

I don't remember seeing `_dependencies.py` but will check it.

synchon: Now implemented with an improved regex search.

PR #9, opened 2022-10-23
[ENH] Implement `junifer selftest` to allow the users to test the library themselves

synchon:
It might be the case that some users have specific working environments that are not comprised within the tests from junifer (i.e. specific versions of the dependencies). For solving issues in a more efficient way, it might be good if we can have a `junifer selftest` command that run the test suite within the environment (so no `tox`, just `pytest`).

### DIFF
--- a/junifer/api/cli.py
+++ b/junifer/api/cli.py
@@ -5,6 +5,9 @@
 # License: AGPL
 
 import pathlib
+import subprocess
+import sys
+from pathlib import Path
 from typing import Dict, List, Union
 
 import click
@@ -216,3 +219,16 @@ def wtf(long_: bool) -> None:
         "environment": _get_environment_information(long_=long_),
     }
     click.echo(yaml.dump(report, sort_keys=False))
+
+
+@cli.command()
+def selftest() -> None:
+    """Selftest command for CLI."""
+    subprocess.run(
+        ["pytest", "-vvv"],
+        stdin=subprocess.DEVNULL,
+        stdout=sys.stdout,
+        stderr=subprocess.STDOUT,
+        cwd=Path(__file__).parent.parent.parent.absolute(),
+        check=True,
+    )
--- a/junifer/api/cli.py
+++ b/junifer/api/cli.py
@@ -222,13 +222,64 @@ def wtf(long_: bool) -> None:
 
 
 @cli.command()
-def selftest() -> None:
-    """Selftest command for CLI."""
-    subprocess.run(
-        ["pytest", "-vvv"],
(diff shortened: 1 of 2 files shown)


synchon:
- The `pytest` invocation is done via `subprocess` as running it from the same process will use the cache after the first run.
- The current state redirects the subprocess `stdout` to `sys.stdout` and hence prints the `pytest` output in "real-time" but of course is dependent on system's buffer.
- The `junifer selftest` command takes ~6-7 minutes to complete, so testing it by invoking in the test suite is a bit tricky now.

fraimondo:
> * The `pytest` invocation is done via `subprocess` as running it from the same process will use the cache after the first run.
> * The current state redirects the subprocess `stdout` to `sys.stdout` and hence prints the `pytest` output in "real-time" but of course is dependent on system's buffer.
> * The `junifer selftest` command takes ~6-7 minutes to complete, so testing it by invoking in the test suite is a bit tricky now.

Was it better to use `subprocess` than just pytest python api? https://docs.pytest.org/en/7.1.x/how-to/usage.html#calling-pytest-from-python-code

synchon:
> > * The `pytest` invocation is done via `subprocess` as running it from the same process will use the cache after the first run.
> 
> > * The current state redirects the subprocess `stdout` to `sys.stdout` and hence prints the `pytest` output in "real-time" but of course is dependent on system's buffer.
> 
> > * The `junifer selftest` command takes ~6-7 minutes to complete, so testing it by invoking in the test suite is a bit tricky now.
> 
> 
> 
> Was it better to use `subprocess` than just pytest python api? https://docs.pytest.org/en/7.1.x/how-to/usage.html#calling-pytest-from-python-code

Yeah it is related to the first point and since we intend to run it multiple times, we need separate process everytime. So, we create new process everytime, "why not just call the command?" was my thought.

fraimondo:
What about using `--cache-clear`?

fraimondo:
https://docs.pytest.org/en/6.2.x/cache.html

synchon:
> What about using `--cache-clear`?

Do you mean to use this along with the subprocess implementation?

fraimondo:
No. I meant to use the python interface with cache clear.

synchon:
So I tested it and it works fine as well. Do we use this then?

fraimondo:
Yes, sure.

synchon:
I only see one problem with this approach which I find after using it multiple times is that it does not handle keyboard interrupt gracefully.

fraimondo (approved):

PR #39, opened 2022-10-24
[DOC]: Provide a thorough description on how junifer works

synchon:
### Which element of the documentation do you want to modify and why?

The documentation lacks a description on how junifer works.

This should include a diagram of the pipeline and what are the different parts of it.

### Anything else to say?

_No response_


synchon:
@fraimondo Maybe we add the diagram later?

fraimondo (approved):

PR #112, opened 2022-11-02
[ENH]: Add support for python 3.11

synchon:

Support Python 3.11

- Test Python 3.11
- Check `raise_error` implementation, as Python 3.11 has a new error reporting feature that is quite cool: https://docs.python.org/3/whatsnew/3.11.html

_No response_

### Anything else to say?

_No response_

### DIFF
--- a/.github/workflows/ci-docs.yml
+++ b/.github/workflows/ci-docs.yml
@@ -1,6 +1,6 @@
 name: Run Tests
 
-on: 
+on:
   push:
     paths-ignore:
       - 'junifer/**'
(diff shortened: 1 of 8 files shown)


synchon:
> Check raise_error implementation, as Python 3.11 has a new error reporting feature that is quite cool: https://docs.python.org/3/whatsnew/3.11.html

Well, `raise_error` is just a simple function which logs error and then raises the exception, so not sure what we can improve here.

fraimondo:
> > Check raise_error implementation, as Python 3.11 has a new error reporting feature that is quite cool: https://docs.python.org/3/whatsnew/3.11.html
> 
> Well, `raise_error` is just a simple function which logs error and then raises the exception, so not sure what we can improve here.

It can be improved, as the traceback is not now optimal.

I also checked how the new Python 3.11 feature works, and I can't find how to use that feature with user raised exceptions.

I'll leave that for a future PR

fraimondo (approved):

PR #102, opened 2022-11-10
[ENH]: Further Aggregation methods for SphereAggregation Marker

synchon:

Requesting this as per:

```python3
if method != "mean":
    raise_error(
        "Only mean aggregation is supported for sphere aggregation. "
        "If you need other aggregation methods, please open an issue "
        "on `junifer github`_.",
        NotImplementedError,
    )
```
in https://github.com/juaml/junifer/blob/main/junifer/markers/sphere_aggregation.py.

I think it would be very useful if one can pass a function or name of function to the SphereAggregation Marker (other than 'mean').
This would be directly relevant for #36, as one could then simply define a kendalls_w function that can be passed as an aggregation method to SphereAggregation, which would give the reho value for that Sphere (if I understand this correctly).

I had a very quick look at nilearn and here is how they do it in a nutshell:
(from: https://github.com/nilearn/nilearn/blob/98a3ee060/nilearn/maskers/nifti_spheres_masker.py#L206)

```python3
for i, sphere in enumerate(_iter_signals_from_spheres(
    self.seeds_, imgs, self.radius, self.allow_overlap,
    mask_img=self.mask_img)):
    signals[:, i] = np.mean(sphere, axis=1)
    return signals, None
```

I guess we should be able to do something similar simply by importing _iter_signals_from_spheres and using a function otther than np.mean() (again, only if I understand this correctly.)

```shell
see above
```

### Anything else to say?

_No response_

### DIFF
--- a/junifer/external/nilearn/junifer_nifti_spheres_masker.py
+++ b/junifer/external/nilearn/junifer_nifti_spheres_masker.py
@@ -1,0 +1,246 @@
+"""Provide custom nilearn.maskers.NiftiSpheresMasker tuned for junifer."""
+
+# Authors: Synchon Mandal <s.mandal@fz-juelich.de>
+# License: AGPL
+
+from typing import TYPE_CHECKING, Callable, List, Tuple, Union
+
+import numpy as np
+from nilearn._utils.class_inspect import get_params
+from nilearn._utils.niimg import img_data_dtype
+from nilearn._utils.niimg_conversions import check_niimg_4d
+from nilearn.maskers import NiftiSpheresMasker
+from nilearn.maskers.base_masker import _filter_and_extract
+from nilearn.maskers.nifti_spheres_masker import _iter_signals_from_spheres
+
+from ...utils import raise_error
+
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from nibabel import Nifti1Image, Nifti2Image
+    from numpy.typing import ArrayLike, DTypeLike
+    from pandas import DataFrame
+
+
+class _JuniferExtractionFunctor:
+    """Functor to extract signals from spheres.
+
+    Parameters
+    ----------
+    seeds_ : list of triple of coordinates in native space
+        Seed definitions. List of coordinates of the seeds in the same space
+        as the images (typically MNI or TAL).
+    radius : float
+        Indicates, in millimeters, the radius for the sphere around the seed.
+        Signal is extracted on a single voxel.
+    mask_img : Niimg-like object
+        Mask to apply to regions before extracting signals.
+    agg_func : callable
+        The function to aggregate signals using.
+    allow_overlap : bool
+        Whether to allow maps to overlap.
+    dtype : any type that can be coerced into a numpy dtype.
+        The dtype for the extraction.
+
+    """
+
+    func_name = "junifer_nifti_spheres_masker_extractor"
+
+    def __init__(
+        self,
+        seeds_: "ArrayLike",
+        radius: float,
+        mask_img: Union["Nifti1Image", "Nifti2Image"],
+        agg_func: Callable,
+        allow_overlap: bool,
+        dtype: "DTypeLike",
+    ) -> None:
+        self.seeds_ = seeds_
+        self.radius = radius
+        self.mask_img = mask_img
+        self.agg_func = agg_func
+        self.allow_overlap = allow_overlap
+        self.dtype = dtype
+
+    def __call__(
+        self,
+        imgs: Union["Nifti1Image", "Nifti2Image"],
+    ) -> Tuple["ArrayLike", None]:
+        """Implement function call overloading.
+
+        Parameters
+        ----------
+         imgs : 4D Niimg-like object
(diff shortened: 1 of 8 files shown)


junifer/external/nilearn/junifer_nifti_spheres_masker.py
```diff
@@ -0,0 +1,255 @@
+"""Provide custom nilearn.maskers.NiftiSpheresMasker tuned for junifer."""
+
+# Authors: Synchon Mandal <s.mandal@fz-juelich.de>
```
fraimondo, line 3: Is this code from Nilearn? If so, can you adjust the license/info on this file to reflect that?

junifer/external/nilearn/tests/test_junifer_nifti_spheres_masker.py
```diff
@@ -0,0 +1,299 @@
+"""Test custom nilearn.maskers.NiftiSpheresMasker."""
+
+# Authors: Synchon Mandal <s.mandal@fz-juelich.de>
```
fraimondo, line 3: Same here, how much is it nilearn's code?

fraimondo (changes requested):

junifer/external/nilearn/junifer_nifti_spheres_masker.py
synchon: Yeah from nilearn, okay will add the information.

junifer/external/nilearn/tests/test_junifer_nifti_spheres_masker.py
synchon: The whole of it. I removed the tests for `inverse_transform`.

fraimondo: then keep nilearn's header then.

junifer/external/nilearn/junifer_nifti_spheres_masker.py
synchon: Okay two doubts:
- nilearn has a different [style](https://github.com/nilearn/nilearn/blob/c3a9390af2349e3263fd0b2991c17a927a142f6c/nilearn/maskers/nifti_spheres_masker.py#L1) of module docstrings, so it will look way different when it renders.
- According to BSD license (which nilearn uses), needs the [license](https://github.com/nilearn/nilearn/blob/main/LICENSE) to be there in anything that redistributes the code.

fraimondo: 1) Change the docstrings
2) Keep the header of the file where it says MIT License, including the authors and stuff

fraimondo: https://github.com/nilearn/nilearn/blob/98a3ee060b55aa073118885d11cc6a1cecd95059/LICENSE#L4-L35

synchon: > * Change the docstrings

You mean make it nilearn style?

> * Keep the header of the file where it says MIT License, including the authors and stuff

They have BSD but I get your point.

junifer/external/nilearn/tests/test_junifer_nifti_spheres_masker.py
synchon: This will also need the license, I believe?

junifer/external/nilearn/junifer_nifti_spheres_masker.py
synchon: > https://github.com/nilearn/nilearn/blob/98a3ee060b55aa073118885d11cc6a1cecd95059/LICENSE#L4-L35

Shouldn't it be the latest license? Licensing is always tricky.

fraimondo: > > * Change the docstrings
> 
> You mean make it nilearn style?

Make them our style




> > https://github.com/nilearn/nilearn/blob/98a3ee060b55aa073118885d11cc6a1cecd95059/LICENSE#L4-L35
> 
> Shouldn't it be the latest license? Licensing is always tricky.

No, the first is nilearn license. The rest are software incorporated in nilearn.

junifer/external/nilearn/tests/test_junifer_nifti_spheres_masker.py
fraimondo: Yes.

junifer/external/nilearn/junifer_nifti_spheres_masker.py
synchon: I meant that the link you referenced has `2007-2015` and the one I referenced has `2007-2022`. So, should we go for the new one?

fraimondo: check the main branch: https://github.com/nilearn/nilearn/blob/main/LICENSE

fraimondo (approved):

PR #110, opened 2022-11-16
[ENH] Include tests for the `queue` command

synchon:
`julearn.api.functions.queue` and `julearn.api.functions._queue_condor` are currently not tested.

The functionality should be tested. At least that the YAML file is created, the submit file is somehow correct, the executable file is there, etc etc.

To consider:
- The logs directory should exist/be created. Otherwise the jobs will be on hold forever.

It will be ideal if we can check the validity of the submit file according to htcondor syntax.

### DIFF
--- a/junifer/api/tests/test_functions.py
+++ b/junifer/api/tests/test_functions.py
@@ -10,6 +10,7 @@ from pathlib import Path
 from typing import List, Tuple, Union
 
 import pytest
+import yaml
 
 import junifer.testing.registry  # noqa: F401
 from junifer.api.functions import collect, queue, run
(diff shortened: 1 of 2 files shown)


junifer/api/tests/test_functions.py
```diff
@@ -144,10 +147,294 @@ def test_run_and_collect(tmp_path: Path) -> None:
...
+        # Create test file, keeping it simple without conditionals
+        (tmp_path / "a.py").touch()
+        with caplog.at_level(logging.DEBUG):
+            queue(
+                config={"with": with_},
+                kind="HTCondor",
+                elements="sub-001",
+            )
+            assert "Copying" in caplog.text
+            assert "Queue done" in caplog.text
```
fraimondo, line 279: Can you assert that the file is present in the jobdir?

junifer/api/tests/test_functions.py
```diff
@@ -144,10 +147,294 @@ def test_run_and_collect(tmp_path: Path) -> None:
...
+
+
+@pytest.mark.parametrize(
+    "elements",
+    [
+        "sub-001",
+        ["sub-001"],
+        ["sub-001", "sub-002"],
+    ],
+)
```
fraimondo, line 289: Can you add elements that are tuples? Like `[("sub-001", "ses-001"),("sub-001", "ses-002")]`

fraimondo (changes requested):
Can you add to the `queue` test a check for the YAML file? It should have all the fields from the original config but the `queue` section.

Can you also add a test for `_queue_condor` checking that the:

* DAG file exists 
* submit file exist,
* that the DAG files has as many run lines as elements
* mem, cpus and disk parameter has effect
* if collect is true, then the DAG should have a job collect. collect submit file should be present anyways
* extra preamble parameter has effect
* Check for submit = True: should give a File not found error or something like that

junifer/api/tests/test_functions.py
synchon: Done.

junifer/api/tests/test_functions.py
synchon: Done.

synchon:
@fraimondo I believe I have addressed your comments.

fraimondo (approved):
Excellent!

PR #135, opened 2022-11-21
Enh/default single output

### LINKED ISSUE #134: [ENH]: Change storage default parameter `single_output` to `True`

Using a Storage requires the user to specify the URI and `single_output=True`, which is counter-intuitive and does not make any sense. 

The only reason why we want `False` is for the `run` and `collect` functions. Because this functions are meant to create one file per subject (`single_output=False`). 

The fix is to set that parameter to `False` in those functions (if not specified by the user) and leave the default as `True`, so it's easier to use the python interface to read features.

Done it, will submit the PR now.

_No response_

### Anything else to say?

_No response_

fraimondo:
* [x] fix #134 
* [x] tests added/passed

### DIFF
--- a/docs/changes/latest.inc
+++ b/docs/changes/latest.inc
@@ -101,3 +101,5 @@ Bugs
 
 API changes
 ~~~~~~~~~~~
+
+- Change the ``single_output`` default parameter in storage classes from ``False`` to ``True`` (:gh:`134` by `Fede Raimondo`_).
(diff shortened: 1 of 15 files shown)


synchon (commented):
Looks good! Can you please run `black` once?

fraimondo:
> Looks good! Can you please run `black` once?

done

synchon (approved):

PR #36, opened 2022-11-24
[MARKER]: ReHo

synchon:

ReHo

Inputs:
List of 3D MNI coordinates and sphere size OR a mask file
A GM brain mask 

Output:
An array with one value per coordinate/voxel

Notes:
With coordinates and sphere size, there are two ways to calculate this:
- use all the voxels in the sphere and calculate Kendall's W
- calculate for each voxel in the sphere using the neighboring voxels

The resulting values are then averaged to get a single value for each coordinate.

https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5021216/

```shell
https://pypi.org/project/kendall-w/
https://stackoverflow.com/questions/48893689/kendalls-coefficient-of-concordance-w-in-python
```

### Anything else to say?

_No response_


junifer/markers/reho_parcels.py
```diff
@@ -0,0 +1,150 @@
...
+            on="BOLD",
+        )
+        # Perform aggregation
+        aggregated_values = parcel_aggregation.compute(
+            input=input, extra_input=extra_input
+        )
+        # Create a new dictionary for returning
+        output = {}
+        # Expand row dimension
+        output["data"] = np.expand_dims(aggregated_values["data"], axis=0)
```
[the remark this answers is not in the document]
fraimondo: I would change `aggregated_values` to `output` and modify in place. Less memory.

junifer/markers/reho_spheres.py
```diff
@@ -0,0 +1,158 @@
...
+        logger.debug("Calculating ReHO for spheres.")
+        # Initialize sphere aggregation
+        sphere_aggregation = SphereAggregation(
+            coords=self.coords,
+            radius=self.radius,
+            method="kendall_w",
+            on="BOLD",
+        )
+        # Perform aggregation
+        aggregated_values = sphere_aggregation.compute(
```
fraimondo, line 147: same comments as before.

junifer/markers/tests/test_reho_parcels.py
```diff
@@ -0,0 +1,66 @@
...
+        reho_parcels_output = reho_parcels_marker.fit_transform(
+            {"BOLD": {"data": fmri_img}}
+        )
+        # Get BOLD output
+        reho_parcels_output_bold = reho_parcels_output["BOLD"]
+        # Assert BOLD output
+        assert "data" in reho_parcels_output_bold
+        assert "columns" in reho_parcels_output_bold
+        assert "rows_col_name" in reho_parcels_output_bold
+        # TODO: add comparison with afni 3dReHo generated data
```
fraimondo, line 37: How difficult is to do this before merging the PR?

junifer/markers/reho_parcels.py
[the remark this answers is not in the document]
synchon: > output["data"] = aggregated_values["data"][np.newaxis, :]

It's the same thing: https://numpy.org/doc/stable/reference/generated/numpy.expand_dims.html
I can change if you like the other one.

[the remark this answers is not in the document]
synchon: > I would change `aggregated_values` to `output` and modify in place. Less memory.

I was trying to not modify in-place but not a bad choice either way, will change.

junifer/markers/tests/test_reho_parcels.py
synchon: I'm working on it but there's some issue with type conversions in afni which is blocking it. And, not sure if other errors will pop up after that.

junifer/markers/reho/reho_estimator.py
```diff
@@ -0,0 +1,551 @@
...
+                mask_cluster[2, 2, 0] = 0
+                mask_cluster[0, 0, 2] = 0
+                mask_cluster[0, 2, 2] = 0
+                mask_cluster[2, 0, 2] = 0
+                mask_cluster[2, 2, 2] = 0
+
+            # Convert 0 / 1 array to bool
+            logical_mask_cluster = mask_cluster.astype(bool)
+
+            for i, j, k in product(
```
fraimondo, line 342: This can be written only once (no need for 125 neigh separately)

Just set:
```
st = 1
end = 2
if nneigh == 125:
   st = 2
   end = 3
```
end then index as `i - st : i + end`, with the product as `range(st, n_x - (end - 1)`

junifer/markers/reho/reho_parcels.py
```diff
@@ -0,0 +1,108 @@
...
+
+    Parameters
+    ----------
+    parcellation : str
+        The name of the parcellation. Check valid options by calling
+        :func:`junifer.data.parcellations.list_parcellations`.
+    use_afni : bool, optional
+        Whether to use AFNI for computing. If None, will use AFNI only
+        if available (default None).
+    reho_params : dict, optional
```
fraimondo, line 29: This needs to be documented.

What are the valid reho params?

junifer/markers/reho/reho_spheres.py
```diff
@@ -0,0 +1,116 @@
...
+        :func:`junifer.data.coordinates.list_coordinates` for options.
+    radius : float, optional
+        The radius of the sphere in millimeters. If None, the signal will be
+        extracted from a single voxel. See
+        :class:`nilearn.maskers.NiftiSpheresMasker` for more information
+        (default None).
+    use_afni : bool, optional
+        Whether to use AFNI for computing. If None, will use AFNI only
+        if available (default None).
+    reho_params : dict, optional
```
fraimondo, line 34: Same as with parcels

junifer/markers/reho/tests/test_reho_parcels.py
```diff
@@ -0,0 +1,112 @@
...
+        # Initialize marker
+        reho_parcels_marker = ReHoParcels(parcellation=PARCELLATION)
+        # Fit transform marker on data
+        reho_parcels_output = reho_parcels_marker.fit_transform(
+            {"BOLD": {"path": "/tmp", "data": fmri_img, "meta": {}}}
+        )
+        # Get BOLD output
+        reho_parcels_output_bold = reho_parcels_output["BOLD"]
+        # Assert BOLD output
+        assert "data" in reho_parcels_output_bold
```
fraimondo, line 37: Things to check:
* shape
* values should be normalized

junifer/markers/reho/tests/test_reho_map_comparison.py
```diff
@@ -0,0 +1,95 @@
...
+            # Compare
+            assert np.testing.assert_allclose(
+                python_reho_map.get_fdata(), afni_reho_map.get_fdata()
+            )
+
+    return ReHoMapComparer
+
+
+@pytest.mark.skip(
+    "should not run due to minor inconsistencies between implementations"
```
fraimondo, line 77: The comparison might not be correct, but we can still test the implementation separately:

Things to check:
* shape
* values should be normalized

fraimondo (changes requested):
There is also a `.nii` file there without a name. I think it should not be there.

junifer/markers/reho/tests/test_reho_map_comparison.py
synchon: I think it's better to leave this module as it is and try to get it tested by having afni in the CI as we had planned.

synchon:
@fraimondo I believe I have addressed your comments.

fraimondo (changes requested):
There are 75 Mb of NII files. We don't need those ones.

Indeed I was wondering if we can just test afni v python only when we have the afni toolbox available and avoid shipping this binaries just for testing.

I do that for ALFF

synchon:
> There are 75 Mb of NII files. We don't need those ones.
> 
> Indeed I was wondering if we can just test afni v python only when we have the afni toolbox available and avoid shipping this binaries just for testing.
> 
> I do that for ALFF

I'm working on getting afni into CI, testing some stuff. I can remove the .nii files and get this in if you like. I'll anyway add the gh-actions stuff in separate PR.

fraimondo:
> > There are 75 Mb of NII files. We don't need those ones.
> > Indeed I was wondering if we can just test afni v python only when we have the afni toolbox available and avoid shipping this binaries just for testing.
> > I do that for ALFF
> 
> I'm working on getting afni into CI, testing some stuff. I can remove the .nii files and get this in if you like. I'll anyway add the gh-actions stuff in separate PR.

That's exactly what I meant! Perfect!

synchon:
> > > There are 75 Mb of NII files. We don't need those ones.
> > > Indeed I was wondering if we can just test afni v python only when we have the afni toolbox available and avoid shipping this binaries just for testing.
> > > I do that for ALFF
> > 
> > 
> > I'm working on getting afni into CI, testing some stuff. I can remove the .nii files and get this in if you like. I'll anyway add the gh-actions stuff in separate PR.
> 
> That's exactly what I meant! Perfect!

I have removed the `.nii` files and adapted the tests.

fraimondo (approved):
Lets merge once the CI is ready to test AFNI stuff

synchon:
Waiting for #153 now.

fraimondo (dismissed):

PR #120, opened 2022-11-24
[DOC]: Fill in built-in coordinates list

synchon:
### Which element of the documentation do you want to modify and why?

Like Parcellations (https://juaml.github.io/junifer/main/builtin.html) but for Coordinates

### Anything else to say?

_No response_

### DIFF
--- a/docs/builtin.rst
+++ b/docs/builtin.rst
@@ -359,13 +359,18 @@ Available
        | The brain's default network: anatomy, function, and relevance to disease.
        | Annals of the New York Academy of Sciences, Volume 1124(1), Pages 1-38 (2008).
        | https://doi.org/10.1196/annals.1440.011
-   * - Extended multiple-demand network
+   * - Missing formal name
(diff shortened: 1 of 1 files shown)


synchon:
We need to anyway update the list later. For now, it has as much information we have available.

synchon:
> @synchon We have 15 coordinates in the `VOIs/meta` directory, why only 9 in the available and 3 in the planned?
> 
> They are all available, problem is that we don't have the right documentation?

Yes, exactly.

fraimondo:
Or either we:

1. Remove for the moment
2. Add in the doc that we don't have the details yet.

synchon:
> Or either we:
> 
> 1. Remove for the moment
> 2. Add in the doc that we don't have the details yet.

2. sounds good to me.

synchon:
I have added as much as I can with some more info.

fraimondo (approved):

PR #145, opened 2022-11-24
[ENH]: Complexity features

### LINKED ISSUE #61: [MARKER]: Add complexity features

Permutation entropy
Range entropy
Multiscale entropy
Hurst exponent

https://github.com/omidvarnia/RangeEn

_No response_

### Anything else to say?

_No response_

omidvarnia:
* [X] fix #61
* [X] description of feature/fix
* [X] tests added/passed
* [X] add an entry for the latest changes

This PR adds seven complexity measures, i.e., range entropy and its AUC, permutation entropy, sample entropy, multiscale entropy and its AUC, and Hurst exponent.


fraimondo:
@omidvarnia Can you create one marker per measure? The way it is right now is one marker to rule them all. Junifer's philosophy is to have specific markers with minimal parametrisation.

docs/changes/latest.inc
```diff
@@ -35,6 +35,8 @@ Enhancements
 - Add :class:`junifer.markers.EdgeCentricFCParcels` and :class:`junifer.markers.EdgeCentricFCSpheres`
   (:gh:`64` by `Leonard Sasse`_).
 
+- Add seven complexity markers (:gh:`61` by `Amir Omidvarnia`_, `Leonard Sasse`_).
```
synchon, line 38: Would be good to have the references to the classes here.

junifer/markers/utils.py
```diff
@@ -5,15 +5,17 @@
 #          Sami Hamdan <s.hamdan@fz-juelich.de>
 #          Synchon Mandal <s.mandal@fz-juelich.de>
 #          Federico Raimondo <f.raimondo@fz-juelich.de>
+#          Amir Omidvarnia <a.omidvarnia@fz-juelich.de>
 # License: AGPL
 
 from typing import Any, Callable, Dict, List, Type, Union
 
+import neurokit2 as nk
```
synchon, line 13: Good to have a `try...catch` here.

pyproject.toml
```diff
@@ -48,6 +48,7 @@ dependencies = [
     "sqlalchemy>=1.4.27,<= 1.5.0",
     "pyyaml>=5.1.2,<7.0",
     "importlib_metadata; python_version < '3.10'",
+    "neurokit2>=0.1.7"
```
synchon, line 51: It's better to have `neurokit2` as optional dependency as it has incompatibilities with `junifer` in some mutual packages like `pandas`.

synchon (changes requested):
@omidvarnia
- The tests are failing for storage and I believe it has something to do with how the marker computation output is being manipulated, would be good if you take a look.
- Would be great if you also take a look at #175 for handling the mask parameter.

docs/changes/latest.inc
omidvarnia: The references to all features were added to latest.inc.

junifer/markers/utils.py
omidvarnia: Sure, we will do it with Fede.

pyproject.toml
omidvarnia: Agreed.

synchon: For the optional dependency block, I have two suggestions which will work like so:
- `pip install junifer[physio]` : to take care of future uses of `neurokit2`
- `pip install junifer[complexity]` : to make it direct

Let me know your thoughts @omidvarnia @fraimondo.

junifer/markers/utils.py
synchon: With the refactored versions, this is not relevant anymore.

pyproject.toml
fraimondo: `neurokit2`

synchon:
Needs to wait for #269, but should be good for review.

junifer/markers/complexity/sample_entropy.py
```diff
@@ -0,0 +1,127 @@
+"""Provide class for sample entropy of a time series."""
+
+# Authors: Amir Omidvarnia <a.omidvarnia@fz-juelich.de>
+#          Leonard Sasse <l.sasse@fz-juelich.de>
+# License: AGPL
+
+from typing import Dict, List, Optional, Union
+
+import neurokit2 as nk
```
fraimondo, line 9: Same here, this makes it mandatory.

junifer/markers/complexity/range_entropy_auc.py
```diff
@@ -0,0 +1,138 @@
+"""Provide class for range entropy of a time series."""
+
+# Authors: Amir Omidvarnia <a.omidvarnia@fz-juelich.de>
+#          Leonard Sasse <l.sasse@fz-juelich.de>
+# License: AGPL
+
+from typing import Dict, List, Optional, Union
+
+import neurokit2 as nk
```
fraimondo, line 9: mandatory import

junifer/markers/complexity/range_entropy.py
```diff
@@ -0,0 +1,129 @@
+"""Provide class for range entropy of a time series."""
+
+# Authors: Amir Omidvarnia <a.omidvarnia@fz-juelich.de>
+#          Leonard Sasse <l.sasse@fz-juelich.de>
+# License: AGPL
+
+from typing import Dict, List, Optional, Union
+
+import neurokit2 as nk
```
fraimondo, line 9: mandatory import

junifer/markers/complexity/perm_entropy.py
```diff
@@ -0,0 +1,125 @@
+"""Provide class for permutation entropy of a time series."""
+
+# Authors: Amir Omidvarnia <a.omidvarnia@fz-juelich.de>
+#          Leonard Sasse <l.sasse@fz-juelich.de>
+# License: AGPL
+
+from typing import Dict, List, Optional, Union
+
+import neurokit2 as nk
```
fraimondo, line 9: mandatory import

junifer/markers/__init__.py
```diff
@@ -23,3 +23,12 @@
     TemporalSNRParcels,
     TemporalSNRSpheres,
 )
+from .complexity import (
```
fraimondo, line 26: To prevent having `neurokit2` as a requirement, I would not import this here.

The idea is that the user specifies `with: junifer.markers.complexity` in the YAML to use this markers.

The other solution would be to use local imports, but that will fail at runtime if the dependencies are not met.

fraimondo (changes requested):

junifer/markers/__init__.py
synchon: So what about importing it automatically if `neurokit2` is found and not importing it and warning when not found? Something like:
```
if importlib.util.find_spec("neurokit2") is None:
    # don't import and just warn
    warn(...)
else:
    # import complexity markers
```

synchon:
@fraimondo I've adapted the imports and everything else as we discussed, should be good for a  review.

docs/api/markers.rst
```diff
@@ -4,3 +4,11 @@ Markers
 .. automodule:: junifer.markers
    :members:
    :imported-members:
+
+
+Complexity
+----------
+
+.. automodule:: junifer.markers.complexity
```
fraimondo, line 12: Will doc build without the neurokit2 dependency? Just asking, I don't know.

fraimondo (approved):

docs/api/markers.rst
synchon: I've added it as a dependency for building docs preview, once that is ok, will add it for building docs.

synchon: Done.

PR #132, opened 2023-01-05
[ENH]: Allow users to select `types` for datagrabbers in order to avoid downloading unnecesary data.

synchon:

Datalad-based datasets such as DataladAOMICID1000, DataladAOMICPIOP1 and DataladAOMICPIOP2 have several data types including BOLD, T1w, DWI, etc.

So far, the types parameter is not exposed to the constructor, os a user that only requires one data type, will be forced to `datalad get` all the data.

easy, as with the `tasks` parameter, expose the `types` one.

_No response_

### Anything else to say?

_No response_


synchon:
> Can we add a test to check what happens when the user passes a type that is not defined in the datagrabber? I guess the `validate_patterns` should fail.

I believe they are there, can you please check the tests with the name like `test_<datagrabber>_incorrect_data_type()`?

fraimondo (approved):

Issue #170, opened 2023-01-05
[ENH]: Obtain subject/image-specific mask for ParcelAggregation by use of a function


LeSasse:
### Are you requiring a new dataset or marker?

- [X] I understand this is not a marker or dataset request

### Which feature do you want to include?

Currently it is possible to give the [junifer ParcelAggregation](https://juaml.github.io/junifer/main/api/markers.html#junifer.markers.ParcelAggregation) in-built binary masks to determine which voxels should be considered for aggregating the data. However, nilearn provides [a few useful functions to obtain image-specific masks](https://nilearn.github.io/dev/modules/masking.html) for different use cases. Similarly, some datasets provide some subject-specific masks (for example fMRIprep provides masks in a format like "sub-{sub}_task-{task}_space-{space}_desc-brain_mask.nii.gz") that may be useful/desirable to include.

### How do you imagine this integrated in junifer?

 It would be very useful to give the ParcelAggregation "mask" parameter a function that returns a mask like this (either by passing a function or by passing a function name as a string similar to passing an aggregation function). This would of course also require another parameter like "mask_params" as a dictionary that can be used to hand over keyword arguments to the masking functions. Similarly, it would be useful to be able to configure this for other parts of the pipeline that use binary masks (for example "clean_img" in the preprocessing module.)

LeSasse:
@fraimondo i think an even bigger underlying issue that I had not thought of before is that the mask used during the cleaning/preprocessing is/can be a different one, from the mask used during marker computation/parcellation which I guess can lead to uncleaned voxels being included in the marker computation.

fraimondo:
Indeed this is an issue that we can't really solve. If the user does not specify a mask, it uses `compute_brain_mask`. If there is another mask in the parcel/sphere/marker/etc which uses _unclean_ voxels, then there's an issue.

This is also the case if you preprocess a BOLD and you do not apply any mask in the Parcel/Sphere aggregation. If the parcel has voxels that are not part of the computed brain mask (`compute_brain_mask`) or the one specified by the user, it will be a mixture of clean and _unclean_ voxels.

I can think of two options for the moment:

1) Set the preprocessor so it cleans voxels inside the mask, but voxels outside the mask are set to NAN. This way they could also be detected further down the stream.

2) Add a "mask" key/value pair in the "BOLD" value of the data object. So the marker can then apply the same mask, or the intersection between masks.

Number 1 looks like a hack. Number 2 will also allow to have subject-specific masks.

[...]

PR #175, opened 2023-01-17
[ENH]: Support for multiple computed masks

fraimondo:

Extracted from comment on #170 by @LeSasse:

> i would probably just take the literal name of the function and for now implement it with these three functions until there is demand for more:
>
> "compute_epi_mask" -> https://nilearn.github.io/dev/modules/generated/nilearn.masking.compute_epi_mask.html#nilearn.masking.compute_epi_mask
>
>"compute_brain_mask" -> https://nilearn.github.io/dev/modules/generated/nilearn.masking.compute_brain_mask.html#nilearn.masking.compute_brain_mask
>
>"compute_background_mask" -> https://nilearn.github.io/dev/modules/generated/nilearn.masking.compute_background_mask.html#nilearn.masking.compute_background_mask
>
>I think most of the parameters for the functions should be possible to use as well, the important ones are strings and numbers I think. In the very least i think, for "compute_brain_mask" you can set the mask_type as a string. There is an optional target affine parameter as well, but this is only required if you also want to do resampling, which is not needed in this particular step.

In a pipeline step.

_No response_

### Anything else to say?

_No response_

### DIFF
--- a/docs/changes/latest.inc
+++ b/docs/changes/latest.inc
@@ -37,8 +37,8 @@ Enhancements
 
 - Change HCP datagrabber tests to decrease CI running time (:gh:`155` by `Fede Raimondo`_).
 
-- Add support for nilearn computed masks (`compute_epi_mask`, `compute_brain_mask`, `compute_background_mask`)
-  (:gh:`175` by `Fede Raimondo`_).
+- Add support for nilearn computed masks (`compute_epi_mask`, `compute_brain_mask`, `compute_background_mask`,
+  `fetch_icbm152_brain_gm_mask`) (:gh:`175` by `Fede Raimondo`_).
 
 Bugs
 ~~~~
--- a/docs/changes/latest.inc
+++ b/docs/changes/latest.inc
@@ -37,6 +37,9 @@ Enhancements
 
 - Change HCP datagrabber tests to decrease CI running time (:gh:`155` by `Fede Raimondo`_).
 
+- Add support for nilearn computed masks (`compute_epi_mask`, `compute_brain_mask`, `compute_background_mask`)
+  (:gh:`175` by `Fede Raimondo`_).
+
 Bugs
 ~~~~
 
--- a/docs/changes/latest.inc
+++ b/docs/changes/latest.inc
@@ -37,8 +37,8 @@ Enhancements
 
 - Change HCP datagrabber tests to decrease CI running time (:gh:`155` by `Fede Raimondo`_).
 
-- Add support for nilearn computed masks (`compute_epi_mask`, `compute_brain_mask`, `compute_background_mask`,
-  `fetch_icbm152_brain_gm_mask`) (:gh:`175` by `Fede Raimondo`_).
+- Add support for nilearn computed masks (``compute_epi_mask``, ``compute_brain_mask``, ``compute_background_mask``,
+  ``fetch_icbm152_brain_gm_mask``) (:gh:`175` by `Fede Raimondo`_).
 
 Bugs
 ~~~~
--- a/junifer/data/masks.py
+++ b/junifer/data/masks.py
@@ -114,6 +114,7 @@ def list_masks() -> List[str]:
 def get_mask(
     name: str,
     target_img: "Nifti1Image",
+    callable_params: Optional[Dict[str, Any]] = None,
 ) -> "Nifti1Image":
     """Get mask, tailored for the target image.
 
@@ -123,7 +124,8 @@ def get_mask(
         The name of the mask.
     target_img : Nifti1Image
         The image to which the mask will be applied.
-
+    callable_params : dict, optional
+        Parameters to pass to the callable mask function (default None).
     Returns
     -------
     Nifti1Image
@@ -133,8 +135,14 @@ def get_mask(
     resolution = np.min(target_img.header.get_zooms()[:3])
     mask_img, _ = load_mask(name, path_only=False, resolution=resolution)
     if callable(mask_img):
-        mask_img = mask_img(target_img)
+        if callable_params is None:
+            callable_params = {}
+        mask_img = mask_img(target_img, **callable_params)
     else:
+        if callable_params is not None:
+            raise_error(
+                "Cannot pass callable_params to a non-callable mask."
+            )
         mask_img = resample_to_img(
             mask_img,
             target_img,
--- a/junifer/data/masks.py
+++ b/junifer/data/masks.py
@@ -45,12 +45,12 @@ data/masks directory. The user can also register their own masks.
 _available_masks: Dict[str, Dict[str, Any]] = {
     "GM_prob0.2": {"family": "Vickery-Patil"},
     "GM_prob0.2_cortex": {"family": "Vickery-Patil"},
-    "compute_brain": {"family": "Callable", "func": compute_brain_mask},
-    "compute_background": {
+    "compute_brain_mask": {"family": "Callable", "func": compute_brain_mask},
+    "compute_backgroun_mask": {
         "family": "Callable",
         "func": compute_background_mask,
     },
-    "compute_epi": {"family": "Callable", "func": compute_epi_mask},
+    "compute_epi_mask": {"family": "Callable", "func": compute_epi_mask},
 }
 
 
@@ -113,7 +113,7 @@ def list_masks() -> List[str]:
 
 def get_mask(
     name: str,
-    target_img: "Nifti1Image",
+    target_data: Dict[str, Any],
     callable_params: Optional[Dict[str, Any]] = None,
 ) -> "Nifti1Image":
     """Get mask, tailored for the target image.
@@ -122,8 +122,9 @@ def get_mask(
     ----------
     name : str
         The name of the mask.
-    target_img : Nifti1Image
-        The image to which the mask will be applied.
+    target_data : Dict[str, Any]
+        The corresponding item of the data object to which the mask will be
+        applied.
     callable_params : dict, optional
         Parameters to pass to the callable mask function (default None).
     Returns
@@ -132,6 +133,7 @@ def get_mask(
         The mask image.
     """
     # Get the min of the voxels sizes and use it as the resolution
+    target_img = target_data["data"]
     resolution = np.min(target_img.header.get_zooms()[:3])
     mask_img, _ = load_mask(name, path_only=False, resolution=resolution)
     if callable(mask_img):
--- a/junifer/data/masks.py
+++ b/junifer/data/masks.py
@@ -37,14 +37,14 @@ if TYPE_CHECKING:
 _masks_path = Path(__file__).parent / "masks"
 
 
-def _fetch_icbm152_brain_gm_mask(target_img, **kwargs):
+def _fetch_icbm152_brain_gm_mask(target_img: "Nifti1Image", **kwargs):
     """Fetch ICBM152 brain mask and resample.
 
     Parameters
     ----------
     target_img : nibabel.Nifti1Image
         The image to which the mask will be resampled.
-    kwargs : dict
+    **kwargs : dict
         Keyword arguments to be passed to
         :func:`nilearn.datasets.fetch_icbm152_brain_gm_mask`.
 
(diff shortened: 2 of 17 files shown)


fraimondo:
@kaurao @LeSasse:

I started working on this (should be finished quite easily). Question now it's about the "default".

What shall we do as "default":
1) Leave it to None (whatever nilearn does will be the default).
2) "compute_brain" (use the MNI152 1mm-resolution template)
3) "compute_epi" (will only work with BOLD)
4) Other suggestions?

LeSasse:
> @kaurao @LeSasse:
> 
> I started working on this (should be finished quite easily). Question now it's about the "default".
> 
> What shall we do as "default":
> 
>     1. Leave it to None (whatever nilearn does will be the default).
> 
>     2. "compute_brain" (use the MNI152 1mm-resolution template)
> 
>     3. "compute_epi" (will only work with BOLD)
> 
>     4. Other suggestions?

In my opinion now (it has shifted over time) the best default is likely the nilearn default, although this in my experience often lead to memory errors for larger BOLD images without a mask. But better than to apply a mask without the user knowing about it or setting it explicitly.

kaurao:
My take is "no masking" as a defaukt with documentation clearly stating that this might run into memory issues.
Then user can of course specifty `compute_epi_mask` etc as they like.

junifer/data/masks.py
```diff
@@ -28,6 +45,12 @@
 _available_masks: Dict[str, Dict[str, Any]] = {
     "GM_prob0.2": {"family": "Vickery-Patil"},
     "GM_prob0.2_cortex": {"family": "Vickery-Patil"},
+    "compute_brain": {"family": "Callable", "func": compute_brain_mask},
```
LeSasse, line 48: I think it would be better to have the actual function name from nilearn (i.e. "compute_brain_mask") rather than an abbreviation of it.

kaurao:
also how about adding support to some `nilearn` internal masks like:
- `nilearn.datasets.fetch_icbm152_brain_gm_mask`
- `nilearn.datasets.load_mni152_brain_mask`

fraimondo:
> also how about adding support to some nilearn internal masks like:

> nilearn.datasets.fetch_icbm152_brain_gm_mask
> nilearn.datasets.load_mni152_brain_mask

The first one can be added. The second is used in `compute_brain_mask`

fraimondo:
So now I have added support for using functions to "compute" masks. What we have to decide now is the API. That is, how users will use this. Before, each mask was just a name. But now we have quite some functions with parameters.

So let's say we want to use the `compute_brain_mask` with `whole-brain` and a threshold of 0.2:

Option A):
```Python
    marker = ParcelAggregation(
        parcellation="Schaefer100x7",
        method="mean",
        name="test_feature",
        on="VBM_GM",
        mask="compute_brain_mask",
        mask_params={"mask_type":"whole-brain", "threshold": 0.2}
    ) 
```

Option B)
```Python
    marker = ParcelAggregation(
        parcellation="Schaefer100x7",
        method="mean",
        name="test_feature",
        on="VBM_GM",
        mask={"compute_brain_mask": {"mask_type":"whole-brain", "threshold": 0.2}}
    ) 
```

While Option A seems more straigthforward, we also need to consider that we will also support _multiple_ masks by combining them (#174):

Then option A might end up like this:

```Python
marker = ParcelAggregation(
    parcellation="Schaefer100x7",
    method="mean",
    name="test_feature",
    on="VBM_GM",
    mask=["compute_brain_mask", "compute_epi_mask"]
    mask_params={
        "compute_brain_mask": {
            "mask_type":"whole-brain",
            "threshold": 0.2
        },
        "compute_epi_mask": {
            "lower_cutoff": 0.3, 
            "upper_cuttoff": 0.8
        }, 
        "threshold": 1  # full intersection
    }
```

Then option 2 would be like this:
```Python
    marker = ParcelAggregation(
        parcellation="Schaefer100x7",
        method="mean",
        name="test_feature",
        on="VBM_GM",
        mask={
            "compute_brain_mask": {
                "mask_type":"whole-brain",
                "threshold": 0.2
            },
            "compute_epi_mask": {
                "lower_cutoff": 0.3, 
                "upper_cuttoff": 0.8
            }, 
            "threshold": 1  # full intersection
        } 
```

The ideal YAML section should be like this:
```yaml
masks:
  - compute_brain_mask:
    - mask_type: "whole-brain"
    - threshold: "0.2"
  - compute_epi_mask:
    - lower_cutoff: 0.3
    - upper_cutoff: 0.8
  - threshold: 1


Now, option 1 will require some translation/interpretation of the YAML before passing the parameter. Option 2 is a direct intepretation of the yaml. However, we will need to make sure that noone tries to register a mask that is named like one of the parameters of `intersect_mask` (`threshold` and `connected`).

What do you think @kaurao , @LeSasse and @synchon ? Ideally, I would like that both python and YAML uses are "understandable"

**Comment by LeSasse** [comment]
@fraimondo I find option two more intuitive, i.e. simply hand over a dictionary with function name as key and function parameters as value.

**Comment by synchon** [comment]
I would also go with option 2 but just rename the parameter name to `masks`

**Comment by fraimondo** [comment]
> @fraimondo I find option two more intuitive, i.e. simply hand over a dictionary with function name as key and function parameters as value. My only question would be in option 2, how does it work with non-function masks, i.e. where there is only a "key" but not a value (as no further parameters are needed). My guess is, if there is a mask from the dataset/datagrabber then it will for option 2 be sth like:
> 
> ```
> {"dataset/datagrabber (whatever the key is)": /path/to/the/mask}
> ```
> 
> Or how would that fit into the API?

That's the issue. It will be a key with None value. This is assuming that `mask` is a Dict.

```Python
    marker = ParcelAggregation(
        parcellation="Schaefer100x7",
        method="mean",
        name="test_feature",
        on="VBM_GM",
        mask={
            "compute_brain_mask": {
                "mask_type":"whole-brain",
                "threshold": 0.2
            },
            "GM_prob0.2": None,
            "threshold": 1  # full intersection
        }
```

The corresponding yaml then, it's not like I posted before (that's a list of dicts).

It will be something like this (notice that I removed the `-`):

```yaml
masks:
  compute_brain_mask:
    - mask_type: "whole-brain"
    - threshold: "0.2"
  compute_epi_mask:
    - lower_cutoff: 0.3
    - upper_cutoff: 0.8
  threshold: 1
  GM_prob0.2: null
```

Which also makes it quite stupid when we only used "named" masks:

```yaml
masks:
  GM_prob0.2: null
```

So to have a nice Yaml, then the `mask` parameter should accept a str or (list of (dict or str)).

That way you can do this:
```yaml
masks:
  - GM_prob0.2
```
or this
```yaml
masks:
  - compute_brain_mask:
    - mask_type: "whole-brain"
    - threshold: "0.2"
  - compute_epi_mask:
    - lower_cutoff: 0.3
    - upper_cutoff: 0.8
  - threshold: 1
  - GM_prob0.2
```

or even this
```yaml
masks2: [GM_prob0.2, inherit]
```

LeSasse:
I think `masks` taking str or (list of (dict or str)). That way, it is very similar to the `scoring` parameter in scikit-learn's GridSearchCV (for example).

LeSasse:
It looks good to me, changes to implement actually handling/combinging the multiple masks, and changes to preprocessing will be done in #174?

junifer/data/masks.py
```diff
@@ -88,11 +146,66 @@ def list_masks() -> List[str]:
...
 
+def get_mask(
+    mask: Union[str, Dict],
+    target_data: Dict[str, Any],
+) -> "Nifti1Image":
+    """Get mask, tailored for the target image.
+
+    Parameters
+    ----------
+    masks : str or dict
```
LeSasse, line 157: this should be str or list of (str or dict) now, no?

junifer/markers/parcel_aggregation.py
```diff
@@ -11,7 +11,7 @@
 from nilearn.maskers import NiftiMasker
```
LeSasse, line 11: same here: in the ParcelAggregation init method should masks not be this should be str or list of (str or dict) now?

junifer/data/masks.py
fraimondo: for the moment this is for one mask at the time.

junifer/markers/parcel_aggregation.py
fraimondo: same as before. This is for computing masks.

junifer/data/tests/test_masks.py
```diff
@@ -151,3 +171,84 @@ def test_vickery_patil() -> None:
...
+        vbm_gm = input["VBM_GM"]
+        vbm_gm_img = vbm_gm["data"]
+        mask = get_mask(mask="identity", target_data=vbm_gm)
+
+        assert_array_equal(mask.get_fdata(), vbm_gm_img.get_fdata())
+
+    del _available_masks["identity"]
+
+
+def test_nilearn_compute_masks() -> None:
```
synchon, line 219: Is it possible to parametrize this test?

docs/changes/latest.inc
```diff
@@ -32,6 +32,9 @@ Enhancements
 - Organize functional connectivity markers in ``junifer.markers.functional_connectivity``
   (:gh:`107` by `Synchon Mandal`_).
 
+- Add support for nilearn computed masks (`compute_epi_mask`, `compute_brain_mask`, `compute_background_mask`,
```
synchon, line 35: Can we have refs for the functions?

synchon (changes requested):
Along with the comments, lint of the code base (isort + black + flake8 checks) would be great.

Edit: Would be cool if you can also add the updated YAML and example in the docs.

docs/changes/latest.inc
fraimondo: Here I meant to add the junifer options, not the nilearn functions.

junifer/data/masks.py
```diff
@@ -28,6 +70,22 @@
 _available_masks: Dict[str, Dict[str, Any]] = {
     "GM_prob0.2": {"family": "Vickery-Patil"},
     "GM_prob0.2_cortex": {"family": "Vickery-Patil"},
+    "compute_brain_mask": {
+        "family": "Callable",
```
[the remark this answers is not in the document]
fraimondo: Because I use it later. I need to know that the family is "Callable". We will also support other functions later on.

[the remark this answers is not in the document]
synchon: Then can we have a `sub-family` which specified that it's from `nilearn` or custom (for example)?

junifer/data/tests/test_masks.py
synchon: If you parametrize the tests, it will essentially run as the structure as you have now but more atomic and cleaner. What do you mean by "computationally inefficient"? For the ones you require extra steps, it might be worth putting it in other functions. IMO, this makes it maintainable and easy for others to approach it.

docs/changes/latest.inc
synchon: What I meant was a map from `junifer` options to `nilearn` functions.

junifer/data/masks.py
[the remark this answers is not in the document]
fraimondo: no need, why shall we add this? the concept of family is to call the function/use internally, not to organise them for the user.

docs/changes/latest.inc
fraimondo: it's in the docs.

junifer/data/masks.py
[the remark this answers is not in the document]
synchon: I don't see a problem either way, my comment was more from a maintenance POV.

docs/changes/latest.inc
synchon: Okay, all good here.

junifer/data/tests/test_masks.py
fraimondo: done

fraimondo:
> * lint of the code base (isort + black + flake8 checks)

Done. I was dealing with a bugbear issue so that's why I did not look at it for the moment.

> update YAML and example in the docs.

No need to change the YAML for the moment. This will happen in #174. We'll need examples with computing masks + multiple masks.

synchon:
> > * lint of the code base (isort + black + flake8 checks)
> 
> Done. I was dealing with a bugbear issue so that's why I did not look at it for the moment.
> 
> > update YAML and example in the docs.
> 
> No need to change the YAML for the moment. This will happen in #174. We'll need examples with computing masks + multiple masks.

Okay then I would just wait for the CI to complete and if the coverage is good, we merge.

synchon:
@fraimondo Can you please do a rebase on `main`?

synchon (dismissed):

synchon (approved):

PR #147, opened 2023-01-27
[ENH]: HDF5 please

synchon:

pretty please!

as storage

_No response_

### Anything else to say?

_No response_


junifer/storage/hdf5.py
```diff
@@ -0,0 +1,849 @@
...
+        of the file will be prefixed with the respective element.
+        If True, will create only one HDF5 file as specified in the
+        ``uri`` and store all the elements in the same file. Concurrent
+        writes should be handled with care (default True).
+    overwrite : bool or "update", optional
+        Whether to overwrite existing file. If True, will overwrite and
+        if "update", will update existing entry or append (default "update").
+    compression : {0-9}, optional
+        Level of gzip compression: 0 (lowest) to 9 (highest) (default 7).
+    **kwargs : dict
```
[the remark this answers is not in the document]
fraimondo: This still needs to be addressed.

junifer/storage/hdf5.py
```diff
@@ -0,0 +1,849 @@
...
+                fname=uri,
+                title=md5,
+                slash="ignore",
+            )
+        except IOError:
+            raise_error(
+                msg=f"HDF5 file not found at: {uri}",
+                klass=IOError,
+            )
+        else:
```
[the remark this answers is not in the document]
fraimondo: still needs to be addressed.

junifer/storage/hdf5.py
```diff
@@ -0,0 +1,954 @@
...
+        # Parameter absence
+        elif not feature_md5 and not feature_name:
+            raise_error(
+                msg=(
+                    "At least one of `feature_name` or `feature_md5` "
+                    "must be specified."
+                )
+            )
+        # Parameter check pass
+        else:
```
fraimondo, line 299: no need for `else` clause.

junifer/storage/hdf5.py
```diff
@@ -0,0 +1,954 @@
...
+            metadata = read_hdf5(
+                fname=str(self.uri.resolve()),  # type: ignore
+                title="meta",
+                slash="ignore",
+            )
+            # Initialize MD5 variable
+            md5: str = ""
+
+            # Consider feature_md5
+            if feature_md5 and not feature_name:
```
fraimondo, line 310: from the above conditions, if `feature_md5` is not none, then `feature_name` is none. Can you simplify the if conditions here?

Indeed the only condition you need to check is to get the feature_md5 from the name.

junifer/storage/hdf5.py
```diff
@@ -0,0 +1,954 @@
...
+                slash="ignore",
+            )
+            reshaped_data = hdf_data["data"]
+
+            # Generate index for the data
+            logger.debug(f"Generating pandas.MultiIndex for {md5} ...")
+            # Create dictionary for aggregating index data
+            element_idx_dict = defaultdict(list)
+
+            if hdf_data["kind"] == "matrix":
```
fraimondo, line 362: The matrix in the DF format should be flattened. Indeed check what we do with matrices before storing in SQLite. This is what needs to happen now, here.

junifer/storage/hdf5.py
```diff
@@ -0,0 +1,954 @@
...
+            # can be replaced with store_metadata() if run on a loop
+            # for the metadata entries from in_storage
+            try:
+                out_metadata = out_storage._read_metadata()
+            except IOError:
+                out_metadata = {}
+            # Update metadata
+            out_metadata.update(in_metadata)
+            # Save metadata
+            out_storage._write_processed_data(
```
fraimondo, line 854: Why writing here? For each on of the 40k files, for each feature, you are triggering a write.

This should be done on the second step of the collect.

junifer/storage/hdf5.py
```diff
@@ -0,0 +1,954 @@
...
+                    if key not in ("data", "element")
+                }
+                # Join the features element for a chunk
+                data_to_write["element"] = reduce(
+                    lambda acc, elem: acc + elem,
+                    [x["element"] for x in stored_data_for_chunk],
+                    [],
+                )
+                # Write data in chunks to avoid memory usage spikes
+                if features_data.ndim == 2:
```
fraimondo, line 917: you can avoid these two if clauses with one list and then convert to tuple.

junifer/storage/tests/test_hdf5.py
```diff
@@ -0,0 +1,980 @@
...
+
+    # List the stored features
+    features = storage.list_features()
+    # Check the MD5
+    assert "BOLD_fc" == features[meta_md5]["name"]
+
+    # Read into dataframe
+    read_df = storage.read_df(feature_md5=meta_md5)
+    # Check shape of dataframe
+    assert read_df.shape == (4, 3)
```
fraimondo, line 431: See my comment on `store_matrix`. We should have one row per sample. In junifer terms, one row per element.

junifer/storage/hdf5.py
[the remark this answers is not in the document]
synchon: Looks like a comment from an earlier review, resolving it.

junifer/storage/hdf5.py
```diff
@@ -0,0 +1,849 @@
...
+        # Write data
+        self._write_processed_data(
+            fname=uri,
+            processed_data=data_to_write,
+            title=meta_md5,
+        )
+
+        logger.info(f"Wrote HDF5 data to: {uri}")
+
+    def store_matrix(
```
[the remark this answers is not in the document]
synchon: From an earlier review maybe, resolving it.

junifer/storage/hdf5.py
```diff
@@ -0,0 +1,849 @@
...
+            for meta in tqdm(in_metadata, desc="feature"):
+                logger.debug(f"Collecting feature MD5: {meta['meta_md5']} ...")
+                # Load data from new instance
+                in_data = in_storage._read_data(md5=meta["meta_md5"])
+
+                logger.info(
+                    f"Updating HDF5 data with data from: {meta['meta_md5']}"
+                )
+                # Save data
+                out_storage._write_processed_data(
```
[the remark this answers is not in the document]
synchon: Looks like an old review, resolving it.

junifer/storage/hdf5.py
synchon: Addressed in `1d220eff`

junifer/storage/hdf5.py
synchon: Sorry I don't understand the reason. I am making it as explicit as possible.

junifer/storage/hdf5.py
synchon: Maybe I'm not getting the rationale here, what do you mean?

junifer/storage/hdf5.py
synchon: Looks like an old review, resolving.

junifer/storage/hdf5.py
synchon: I know there's a bit of redundancy but kept it for explicit code. What do you mean exactly?

junifer/storage/hdf5.py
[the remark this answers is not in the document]
synchon: We have `**kwargs` in the constructor signature for smooth subclassing and just makes it transparent to the reader that it has it. I don't have a preference here, but I have always found it in the wild and is natural to me at this point. If you have a strong reason, I don't mind.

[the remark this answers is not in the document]
fraimondo: If new parameters arise, then we can change the classes. We are mainly focused on user-oriented documentation, so I want to avoid having things that we don't document / are difficult to follow from the end-user point of view.

junifer/storage/hdf5.py
fraimondo: I meant that before this point, you already checked that only one of `feature_md5` or `feature_name` are not none. There's no need for the `and not feature_name` here.

junifer/storage/hdf5.py
fraimondo: When you convert a matrix to dataframe, it must generate one row per element. As a general rule, each DF must have only one row per element (except for timeseries).

junifer/storage/hdf5.py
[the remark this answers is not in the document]
synchon: Do you mean to remove it from the signature as well as docstring or just the docstring?

junifer/storage/hdf5.py
synchon: Addressed in `25643d91`.

junifer/storage/hdf5.py
synchon: Should be addressed in the latest commits.

junifer/storage/tests/test_hdf5.py
synchon: Addressed.

junifer/storage/hdf5.py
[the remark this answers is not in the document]
fraimondo: nono, it is still unresolved.

Why do you have an else statement? Do you expect other kind of error?

junifer/storage/hdf5.py
```diff
@@ -0,0 +1,959 @@
...
+        if hdf_data["kind"] == "matrix":
+            # Get row and column count
+            n_rows, n_cols = hdf_data["data"][:, :, 0].shape
+            # Set rows for the index
+            for element in hdf_data["element"]:
+                for key, val in element.items():
+                    # Only one row per element
+                    element_idx_dict[key].append(val)
+            # Set column headers for dataframe
+            columns = list(
```
fraimondo, line 370: Check what we do with the SQLIteFeatureStorage. It's not that simple as creating the product, because we might only want to to store a triangular matrix.

Indeed, the code from [here](https://github.com/juaml/junifer/blob/7acc0cd188d44bce4248504ce498e6d55256f945/junifer/storage/sqlite.py#L434) can be extracted to a function in storage.utils so an upper/lower/full matrix can be converted to row in the same way, in both places.

fraimondo (changes requested):

junifer/storage/hdf5.py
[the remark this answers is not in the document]
synchon: Addressed in `18643864`.

junifer/storage/hdf5.py
synchon: The `ndim` would be 3 here but I get your point. Addressed in `07c769e4`.

junifer/storage/hdf5.py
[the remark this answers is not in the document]
synchon: `else` in a `try...except...else` is used for the condition where no exceptions are raised and the block passes.

junifer/storage/hdf5.py
synchon: Addressed with latest commits.

fraimondo (approved):

junifer/storage/hdf5.py
```diff
@@ -0,0 +1,926 @@
...
+            # Flatten data and get column headers for dataframe
+            flat_data, columns = matrix_to_vector(
+                data=hdf_data["data"],
+                col_names=hdf_data["column_headers"],
+                row_names=hdf_data["row_headers"],
+                matrix_kind=hdf_data["matrix_kind"],
+                diagonal=bool(hdf_data["diagonal"]),
+            )
+            # Convert data to proper 2D
+            reshaped_data = flat_data.T
```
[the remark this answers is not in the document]
synchon: Works.

junifer/storage/hdf5.py
```diff
@@ -0,0 +1,926 @@
...
+        logger.debug(f"Generating pandas.MultiIndex for {md5} ...")
+        # Create dictionary for aggregating index data
+        element_idx_dict = defaultdict(list)
+
+        if hdf_data["kind"] == "matrix":
+            # Set rows for the index
+            for element in hdf_data["element"]:
+                for key, val in element.items():
+                    # Only one row per element
+                    element_idx_dict[key].append(val)
```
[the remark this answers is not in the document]
synchon: Addressed.

junifer/storage/hdf5.py
```diff
@@ -0,0 +1,926 @@
...
+                col_names=hdf_data["column_headers"],
+                row_names=hdf_data["row_headers"],
+                matrix_kind=hdf_data["matrix_kind"],
+                diagonal=bool(hdf_data["diagonal"]),
+            )
+            # Convert data to proper 2D
+            reshaped_data = flat_data.T
+        elif hdf_data["kind"] == "vector":
+            # Set rows for the index
+            for element in hdf_data["element"]:
```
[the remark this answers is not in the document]
synchon: Addressed.

PR #190, opened 2023-03-15
Add allow_overlap and allow empty spheres in sphere aggregation

fraimondo:
Features:
* Allow for empty spheres (this will result in arrays with the second dimention of size 0). Applying any aggregation function should yield a NaN.
* Expose `allow_overlap` parameter to every sphere-based marker.

* [x] tests added/passed

### DIFF
--- a/docs/changes/latest.inc
+++ b/docs/changes/latest.inc
@@ -55,6 +55,8 @@ Enhancements
 
 - Expose `allow_overlap` parameter in  :class:`junifer.markers.SphereAgregation` and related markers (:gh:`190` by `Fede Raimondo`_).
 
+- Add aggregation function :func:`junifer.stats.count` that returns the number of elements in a given axis. This allows to count the number of voxels per sphere/parcel when used as `agg_func` in markers (:gh:`190` by `Fede Raimondo`_).
+
 Bugs
 ~~~~
 
--- a/docs/changes/latest.inc
+++ b/docs/changes/latest.inc
@@ -51,6 +51,10 @@ Enhancements
 
 - Add ``pre_run`` parameter to ``_queue_condor`` (:gh:`188` by `Fede Raimondo`_).
 
+- Allow for empty spheres in :class:`junifer.external.nilearn.JuniferNiftiSpheresMasker`, that will result in NaNs (:gh:`190` by `Fede Raimondo`_).
+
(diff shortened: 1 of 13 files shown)


docs/changes/latest.inc
```diff
@@ -51,6 +51,12 @@ Enhancements
 
 - Add ``pre_run`` parameter to ``_queue_condor`` (:gh:`188` by `Fede Raimondo`_).
 
+- Allow for empty spheres in :class:`junifer.external.nilearn.JuniferNiftiSpheresMasker`, that will result in NaNs (:gh:`190` by `Fede Raimondo`_).
+
+- Expose `allow_overlap` parameter in  :class:`junifer.markers.SphereAgregation` and related markers (:gh:`190` by `Fede Raimondo`_).
```
synchon, line 56:
```
``allow_overlap``
```

junifer/external/nilearn/junifer_nifti_spheres_masker.py
```diff
@@ -59,6 +64,155 @@
 """
 
 
+def _apply_mask_and_get_affinity(
```
synchon, line 67: This is for adding the custom check of empty spheres right?

junifer/stats.py
```diff
@@ -82,6 +91,24 @@ def get_aggfunc_by_name(
...
+    Parameters
+    ----------
+    data : numpy.ndarray
+        Data to count elements on.
+    axis : int, optional
+        The axis to count elements on (default 0).
+
+    Returns
+    -------
+    numpy.ndarray
```
synchon, line 106: Should be an `int`.

synchon (changes requested):

junifer/external/nilearn/junifer_nifti_spheres_masker.py
fraimondo: Indeed the only change from nilearn was to remove the check.

fraimondo:
done
can't test more without writing nilearn tests

junifer/external/nilearn/junifer_nifti_spheres_masker.py
synchon: Sounds good.

synchon (approved):

PR #192, opened 2023-03-15
[BUG] Set datalad cache to a temporary location to avoid locks on high throughput settings

fraimondo:
The default for the datalad cache is in the user's home folder. When using throwaway clones with datalad datagrabbers (`datadir=None`) we still have a bottleneck on the cache and the locks, depending on the user-configuration.

The solution is to use the `datalad.cfg.set` with `override=True` to set the cache, locks and sockets directories when the junifer uses throwaway clones. That is, when `datadir is None` and a temporary directory is created.

Additionally, this PR makes sure that the `tempdir` is deleted on exit.

### DIFF
--- a/junifer/datagrabber/datalad_base.py
+++ b/junifer/datagrabber/datalad_base.py
@@ -5,10 +5,13 @@
 #          Synchon Mandal <s.mandal@fz-juelich.de>
 # License: AGPL
 
+import atexit
+import shutil
 import tempfile
 from pathlib import Path
 from typing import Dict, Optional, Tuple, Union
 
+import datalad
 import datalad.api as dl
 from datalad.support.gitrepo import GitRepo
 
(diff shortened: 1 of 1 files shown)


junifer/datagrabber/datalad_base.py
```diff
@@ -85,6 +119,17 @@ def __init__(
         # dirty
         self.datalad_dirty = False
 
+    def __del__(self):
```
synchon, line 122: `def __del__(self) -> None:`

junifer/datagrabber/datalad_base.py
```diff
@@ -85,6 +119,17 @@ def __init__(
...
+        """Destructor."""
+        if hasattr(self, "_tmpdir"):
+            self._rmtmpdir()
+
+    def _rmtmpdir(self):
```
synchon, line 127: `def _rmtmpdir(self) -> None:`

junifer/datagrabber/datalad_base.py
synchon, line 1: Also needs a `black` run for the file.

fraimondo:
done
let's wait for the CI

synchon:
Somehow `black` didn't work before.

synchon (approved):

fraimondo:
It's black in juseless that apparently does not do as you expect.

PR #194, opened 2023-03-16
[BUG]: allow for empty parcels in ParcelAggregation

fraimondo:
Currently, subjects in which a parcel has 0 voxels will result in different sized vectors/matrix/timeseries, which will later fail on collect.

As with spheres, we should allow for "empty" parcels and compute the corresponding NAN value.

### DIFF
--- a/docs/changes/latest.inc
+++ b/docs/changes/latest.inc
@@ -59,6 +59,7 @@ Enhancements
 
 - Modify the behaviour of the ``collect`` parameter in HTCondor ``queue`` function to run a collect job even if some of the previous jobs fail. This is useful to collect the results of a pipeline even if some of the jobs fail (:gh:`190` by `Fede Raimondo`_).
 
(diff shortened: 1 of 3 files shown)


docs/changes/latest.inc
```diff
@@ -69,6 +71,8 @@ Bugs
 
 - Fix an issue with datalad cache and locks in which the user-specific configuration might create a conflict in high throughput systems (:gh:`192` by `Fede Raimondo`_).
 
+- Fix a bug in which :class:`junifer.markers.ParcelAggregation` could yield duplicated colum names if two or more parcels were used and label names were not unique (:gh:`194` by `Fede Raimondo`_).
```
synchon, line 74: ```... column names ...```

synchon (changes requested):

synchon (approved):

synchon (approved):

PR #199, opened 2023-03-20
Fix/datalad cache

fraimondo:
Fix an issue with datalad cache/locks on independent clones (temporary directories)

The main issue is a scalability issue and the cache/locks set in the home folder. At some point, there are way too many processes looking to get the lock, and even with the retries and waiting periods, they just give up and start failing.

Solution: use os.environ["DATALAD_LOCATIONS_LOCKS"] = ... to set an override (so each process has an independent lock/cache) and then datalad.cfg.reload() to force reload the configuration. This is supposed to propagate the override to subproccesses.

Also, this PR allows to use numbers for verbosity levels.

### DIFF
--- a/junifer/api/cli.py
+++ b/junifer/api/cli.py
@@ -93,18 +93,23 @@ def _validate_verbose(
     str or int
         The validated value.
     """
-    valid_values = ["error", "warning", "info", "debug"]
     if isinstance(value, int):
         return value
-    elif isinstance(value, str) and value.lower() in valid_values:
+
+    valid_strings = ["error", "warning", "info", "debug"]
+    if isinstance(value, str) and value.lower() in valid_strings:
         return value.upper()
+
     try:
         value = int(value)  # type: ignore
         return value
     except ValueError:
+        # If we get here, the value is not a valid integer.
         pass
+
+    # If we get here, the value is not valid.
     raise click.BadParameter(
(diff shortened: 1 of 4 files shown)


junifer/api/cli.py
```diff
@@ -74,6 +74,39 @@ def _parse_elements(element: str, config: Dict) -> Union[List, None]:
...
+    if isinstance(value, int):
+        return value
+    elif isinstance(value, str) and value.lower() in valid_values:
+        return value.upper()
+    else:
+        try:
+            value = int(value)  # type: ignore
+            return value
+        except ValueError:
+            pass
```
synchon, line 104: I wouldn't go with a `pass` as it might have unforeseen cases. Better to log it or warn?

synchon (changes requested):

junifer/api/cli.py
fraimondo: It's just an invalid string value that is not a valid option and can't be casted to int. We need to continue and then raise the BadParameter. There's no reason to log/warn here. It's a dead-end.

synchon: I would put the `return` in an `else` block then.

fraimondo: I'll add the comment, but I do not like the `else` block to contain just the return as it complicates reading.

synchon (approved):

PR #197, opened 2023-03-20
[ENH]: Allow to queue a collect job that will run even if some of the previous jobs fail in HTCondor

fraimondo:

Currently, we are using a DAG to do "run" on all elements and finally do a "collect. The main issue with this DAG is that if any of the jobs fail, the collect is not run. With small datasets, this is not a problem. But it is unrealistic to expect this on big datasets.

This leads to manually needing to queueing the collect job later.

Fortunately, DAGman allows to set a "FINAL" jobs which runs once at the end of the dag: https://htcondor.readthedocs.io/en/latest/users-manual/dagman-workflows.html#id2

Change the `collect` argument from `_queue_condor` from boolean to three options:

1) No: will not collect
2) Yes: will collect if all previous jobs were succesfull
3) Always: will collect no matter what happened before

It also needs a new script to be there, so if the users removes the DAG using condor_rm, then the collect job is not run.

the script must be this one (according to the doc), named `collect_pre.pl`. Maybe we can do the same in bash.

```perl
#!/usr/bin/env perl

if ($ARGV[0] eq 4) {
    exit(1);
}
```

```python
        if collect is True:
            dag_file.write(f"JOB collect {submit_collect_fname}\n")
            dag_file.write("PARENT ")
            for i_job, _t_elem in enumerate(elements):
                dag_file.write(f"run{i_job} ")
            dag_file.write("CHILD collect\n\n")
        elif collect is "always":
            dag_file.write(f"FINAL collect {submit_collect_fname}\n")
            dag_file.write(f"SCRIPT PRE collect collect_pre.pl $DAG_STATUS\n")
```

### Anything else to say?

_No response_

### DIFF
--- a/junifer/api/functions.py
+++ b/junifer/api/functions.py
@@ -327,7 +327,7 @@ def _queue_condor(
     extra_preamble: str = "",
     pre_run: Optional[str] = None,
     verbose: str = "info",
-    collect: bool = True,
+    collect: Union[str, bool] = True,
     submit: bool = False,
 ) -> None:
     """Submit job to HTCondor.
@@ -358,8 +358,18 @@ def _queue_condor(
         Extra bash commands to source before the job (default None).
     verbose : str, optional
         The level of verbosity (default "info").
-    collect : bool, optional
+    collect : bool or str, optional
         Whether to submit "collect" task for junifer (default True).
+        Valid options are:
+
+            * True: Submit "collect" task and run even if some of the jobs
+                fail.
+            * "on_success_only": Submit "collect" task and run only if all jobs
+                succeed.
+            * False: Do not submit "collect" task.
+            * "yes": Alias for True.
+             "no": Alias for False.
+
     submit : bool, optional
         Whether to submit the jobs. In any case, .dag files will be created
         for submission (default False).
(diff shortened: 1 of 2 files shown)


junifer/api/functions.py
```diff
@@ -358,8 +358,16 @@ def _queue_condor(
         Extra bash commands to source before the job (default None).
     verbose : str, optional
         The level of verbosity (default "info").
-    collect : bool, optional
+    collect : str, optional
         Whether to submit "collect" task for junifer (default True).
```
synchon, line 362: ```(default "yes")```

junifer/api/functions.py
```diff
@@ -379,6 +387,13 @@ def _queue_condor(
         f"collect {str(yaml_config.absolute())} --verbose {verbose} "
     )
 
+    if not isinstance(collect, str):
+        raise_error("collect must be a string")
+
+    collect = collect.lower()
+    if collect not in ["yes", "no", "on_success_only"]:
+        raise ValueError(f"Invalid value for collect: {collect}")
```
synchon, line 395: `raise_error` instead of `raise`?

junifer/api/tests/test_functions.py
```diff
@@ -723,20 +731,34 @@ def test_queue_condor_assets_generation(
...
                 for line in f.read().splitlines():
                     if "JOB" in line:
                         element_count += 1
                     if "collect" in line:
                         has_collect_job = True
+                    if "FINAL" in line:
+                        has_final_collect_job = True
 
-            if collect:
+            if collect in ["yes", True]:
```
synchon, line 744: ```if collect == "yes":```

synchon (changes requested):

synchon (approved):

PR #213, opened 2023-03-28
[ENH] Use and document `towncrier` for creating and maintaining changelogs

synchon:
* [ ] fix #(issue number)
* [x] tests added/passed

This PR adds support for `towncrier` to maintain changelogs.


fraimondo:
We are missing some things:

1) `make local` should build the docs, including the latest news fragments. This is also used in the pr-preview. 
2) As a consequence of 1, `towncrier` should be a dependency for `docs`
3) `make html` should also build the latest docs, including the news fragment of the "main" branch.

Solution 1) We can, of course, call `towncrier build --keep` to update `whats_new.rst` before doing the `local` and `html`. However, this will modify the `whats_new.rst` file, and this file will be a modified one for git.

Solution 2) We can have a `towncrier_dev.toml` that instead writes to `dev_whats_new.rst`, but then we need to deal with this file when it does not exist (e.g. on release)

Solution 3) Before `make local` and `make html` we create a copy of `whats_new.rst`. We then build the docs and after, we restore the original version.

@synchon: what's your take on this one?

synchon:
> We are missing some things:
> 
> 1. `make local` should build the docs, including the latest news fragments. This is also used in the pr-preview.
> 2. As a consequence of 1, `towncrier` should be a dependency for `docs`
> 3. `make html` should also build the latest docs, including the news fragment of the "main" branch.
> 
> Solution 1) We can, of course, call `towncrier build --keep` to update `whats_new.rst` before doing the `local` and `html`. However, this will modify the `whats_new.rst` file, and this file will be a modified one for git.
> 
> Solution 2) We can have a `towncrier_dev.toml` that instead writes to `dev_whats_new.rst`, but then we need to deal with this file when it does not exist (e.g. on release)
> 
> Solution 3) Before `make local` and `make html` we create a copy of `whats_new.rst`. We then build the docs as in solution 1. After, we restore the original version of `whats_new.rst`.
> 
> @synchon: what's your take on this one?

My take would be:

#. PR-preview: Run the `towncrier` with `--yes` and `--keep` (or not) so it runs without prompt and then build the docs. This would make a `whats_new.rst` staged for commit but we don't bother as long as it builds ok as it's preview.
#. Pre-releases (non-release PRs merged to `main`): Although we update the changelog now, I'm not sure if it makes sense to update the changelog with every pre-release changes. So, I wouldn't make any changes to `whats_new.rst`.
#. Release: We will anyway create a new PR and check everything manually before making a release.

fraimondo:
>

fraimondo:
> > We are missing some things:
> > 
> > 1. `make local` should build the docs, including the latest news fragments. This is also used in the pr-preview.
> > 2. As a consequence of 1, `towncrier` should be a dependency for `docs`
> > 3. `make html` should also build the latest docs, including the news fragment of the "main" branch.
> > 
> > Solution 1) We can, of course, call `towncrier build --keep` to update `whats_new.rst` before doing the `local` and `html`. However, this will modify the `whats_new.rst` file, and this file will be a modified one for git.
> > Solution 2) We can have a `towncrier_dev.toml` that instead writes to `dev_whats_new.rst`, but then we need to deal with this file when it does not exist (e.g. on release)
> > Solution 3) Before `make local` and `make html` we create a copy of `whats_new.rst`. We then build the docs as in solution 1. After, we restore the original version of `whats_new.rst`.
> > @synchon: what's your take on this one?
> 
> My take would be:
> 
> 1. PR-preview: Run the `towncrier` with `--yes` and `--keep` (or not) so it runs without prompt and then build the docs. This would make a `whats_new.rst` staged for commit but we don't bother as long as it builds ok as it's preview.
> 2. Pre-releases (non-release PRs merged to `main`): Although we update the changelog now, I'm not sure if it makes sense to update the changelog with every pre-release changes. So, I wouldn't make any changes to `whats_new.rst`.
> 3. Release: We will anyway create a new PR and check everything manually before making a release.


First reply:

1) PR-preview is not an issue. The problem is the user building the documentation. This will stage changes in git. I don't want that to happen. PR-preview relies on `make local`
2) Pre-releases. The repo should not change. The docs should. Also, it relies on `make html`
3) Release, not an issue, we do it manually

So both of the issues will be solved if the Makefile accounts for this.

synchon:
> > > We are missing some things:
> > > 
> > > 1. `make local` should build the docs, including the latest news fragments. This is also used in the pr-preview.
> > > 2. As a consequence of 1, `towncrier` should be a dependency for `docs`
> > > 3. `make html` should also build the latest docs, including the news fragment of the "main" branch.
> > > 
> > > Solution 1) We can, of course, call `towncrier build --keep` to update `whats_new.rst` before doing the `local` and `html`. However, this will modify the `whats_new.rst` file, and this file will be a modified one for git.
> > > Solution 2) We can have a `towncrier_dev.toml` that instead writes to `dev_whats_new.rst`, but then we need to deal with this file when it does not exist (e.g. on release)
> > > Solution 3) Before `make local` and `make html` we create a copy of `whats_new.rst`. We then build the docs as in solution 1. After, we restore the original version of `whats_new.rst`.
> > > @synchon: what's your take on this one?
> > 
> > 
> > My take would be:
> > 
> > 1. PR-preview: Run the `towncrier` with `--yes` and `--keep` (or not) so it runs without prompt and then build the docs. This would make a `whats_new.rst` staged for commit but we don't bother as long as it builds ok as it's preview.
> > 2. Pre-releases (non-release PRs merged to `main`): Although we update the changelog now, I'm not sure if it makes sense to update the changelog with every pre-release changes. So, I wouldn't make any changes to `whats_new.rst`.
> > 3. Release: We will anyway create a new PR and check everything manually before making a release.
> 
> First reply:
> 
> 1. PR-preview is not an issue. The problem is the user building the documentation. This will stage changes in git. I don't want that to happen. PR-preview relies on `make local`
> 2. Pre-releases. The repo should not change. The docs should. Also, it relies on `make html`
> 3. Release, not an issue, we do it manually
> 
> So both of the issues will be solved if the Makefile accounts for this.

1. When the user builds the documentation, `towncrier` doesn't need to be called so it wouldn't stage anything in git.
2. What if we don't change the docs as we would already have the news fragmnets we need to make the complete doc when making a release. (Building non-release version docs would have been better if we had a `dev` branch which could in turn have a `what_new_dev.rst` but that's a different case.)

fraimondo:
> > > > We are missing some things:
> > > > 
> > > > 1. `make local` should build the docs, including the latest news fragments. This is also used in the pr-preview.
> > > > 2. As a consequence of 1, `towncrier` should be a dependency for `docs`
> > > > 3. `make html` should also build the latest docs, including the news fragment of the "main" branch.
> > > > 
> > > > Solution 1) We can, of course, call `towncrier build --keep` to update `whats_new.rst` before doing the `local` and `html`. However, this will modify the `whats_new.rst` file, and this file will be a modified one for git.
> > > > Solution 2) We can have a `towncrier_dev.toml` that instead writes to `dev_whats_new.rst`, but then we need to deal with this file when it does not exist (e.g. on release)
> > > > Solution 3) Before `make local` and `make html` we create a copy of `whats_new.rst`. We then build the docs as in solution 1. After, we restore the original version of `whats_new.rst`.
> > > > @synchon: what's your take on this one?
> > > 
> > > 
> > > My take would be:
> > > 
> > > 1. PR-preview: Run the `towncrier` with `--yes` and `--keep` (or not) so it runs without prompt and then build the docs. This would make a `whats_new.rst` staged for commit but we don't bother as long as it builds ok as it's preview.
> > > 2. Pre-releases (non-release PRs merged to `main`): Although we update the changelog now, I'm not sure if it makes sense to update the changelog with every pre-release changes. So, I wouldn't make any changes to `whats_new.rst`.
> > > 3. Release: We will anyway create a new PR and check everything manually before making a release.
> > 
> > 
> > First reply:
> > 
> > 1. PR-preview is not an issue. The problem is the user building the documentation. This will stage changes in git. I don't want that to happen. PR-preview relies on `make local`
> > 2. Pre-releases. The repo should not change. The docs should. Also, it relies on `make html`
> > 3. Release, not an issue, we do it manually
> > 
> > So both of the issues will be solved if the Makefile accounts for this.
> 
> 1. When the user builds the documentation, `towncrier` doesn't need to be called so it wouldn't stage anything in git.

Yes they need to do it. Otherwise it would not be the same doc as in the pr-preview / main branch. The dev should be able to build the same docs locally.

> 2. What if we don't change the docs as we would already have the news fragmnets we need to make the complete doc when making a release. (Building non-release version docs would have been better if we had a `dev` branch which could in turn have a `what_new_dev.rst` but that's a different case.)

Pre-release docs (as well as pr-preview) should include the changes in the DEV version. like we have now:
Screenshot 2023-03-29 at 13 07 19

The point is that the `whats_new_rst` file is only commited to git on the release process. But it needs to be modified for:
`make local`, `make html`, which will affect PR-preview and pre-release docs repectively.

fraimondo:
In short, behaviour should not change with respect to what we have now. Only implementation.

synchon:
In that case I would adapt your solutions like this:
- Make a `towncrier_dev.toml` with the difference being in `filename = "docs/whats_new_dev.rst"
- Include `towncrier --keep` in Makefile before `sphinx` and ask user to commit that as well.
- Make a new file `changelog.rst` and include `whats_new.rst` and `whats_new_dev.rst` in there as two sections in it.

fraimondo:
> In that case I would adapt your solutions like this:
> 
> * Make a `towncrier_dev.toml` with the difference being in `filename = "docs/whats_new_dev.rst"
> * Include `towncrier build --config towncrier_dev.toml --keep` in Makefile before `sphinx` and ask user to commit that as well.
> * Make a new file `changelog.rst` and include `whats_new.rst` and `whats_new_dev.rst` in there as two sections in it.

Why `commit`? There's nothing to commit. The ".dev" changes of the whats_new should be re-created every time without commit.

This also does not work, as sphinx complains that `whats_new_dev.rst` does not exist (on releases). If we add an empty file, then we are in the same issue as before (git changes).

Solution: `make local` and `make html` should:

1) towncrier build --keep
2) make the respective target
3) revert `whats_new.rst` to how it was before 1): `git stash -- docs/whats_new.rst`

fraimondo:
needs this in the CI for docs and docs preview:
Otherwise it does not install the full package with history, but only the last commit.

```
      with:
        fetch-depth: 0
        submodules: true
```

fraimondo (dismissed):

PR #223, opened 2023-04-24
Use `ruamel.yaml` in place of `pyyaml`

synchon:
`pyyaml` is lacking two features which we are interested in:

- Support for YAML 1.2
- Multiline string

`ruaml.yaml` provides support for them, so will be worth using it over `pyyaml`.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -23,7 +23,7 @@ The following steps are specific to VSCode and you can choose to go with it:
 
    .. code-block:: bash
 
-       conda env create -n <your-environment-name> -f conda-env.yml python=3.10
+       conda env create -n <your-environment-name> -f conda-env.yml
        conda activate <your-environment-name>
 
    The ``conda-env.yml`` can be found at the root of the repository.
--- a/junifer/api/functions.py
+++ b/junifer/api/functions.py
@@ -12,8 +12,6 @@ import typing
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple, Union
 
-import yaml
-
 from ..datagrabber.base import BaseDataGrabber
 from ..markers.base import BaseMarker
 from ..markers.collection import MarkerCollection
@@ -22,6 +20,7 @@ from ..preprocess.base import BasePreprocessor
 from ..storage.base import BaseFeatureStorage
 from ..utils import logger, raise_error
 from ..utils.fs import make_executable
+from .utils import yaml
 
 
 def _get_datagrabber(datagrabber_config: Dict) -> BaseDataGrabber:
@@ -270,8 +269,7 @@ def queue(
 
     yaml_config = jobdir / "config.yaml"
     logger.info(f"Writing YAML config to {str(yaml_config.absolute())}")
-    with open(yaml_config, "w") as f:
-        f.write(yaml.dump(config))
+    yaml.dump(config, stream=yaml_config)
 
     # Get list of elements
     if elements is None:
--- a/junifer/api/parser.py
+++ b/junifer/api/parser.py
@@ -10,9 +10,8 @@ import sys
 from pathlib import Path
 from typing import Dict, Union
 
-import yaml
-
 from ..utils.logging import logger, raise_error
+from .utils import yaml
 
 
 def parse_yaml(filepath: Union[str, Path]) -> Dict:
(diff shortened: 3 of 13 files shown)


docs/faq.rst
```diff
@@ -23,7 +23,7 @@ The following steps are specific to VSCode and you can choose to go with it:
 
    .. code-block:: bash
 
-       conda env create -n <your-environment-name> -f conda-env.yml python=3.10
```
LeSasse, line 26: What is the reason for not having to specify the python version in the command anymore?

junifer/api/functions.py
```diff
@@ -270,8 +269,7 @@ def queue(
 
     yaml_config = jobdir / "config.yaml"
     logger.info(f"Writing YAML config to {str(yaml_config.absolute())}")
-    with open(yaml_config, "w") as f:
-        f.write(yaml.dump(config))
+    yaml.dump(config, stream=yaml_config)
```
LeSasse, line 272: Why does this not need a context manager anymore?

junifer/api/parser.py
```diff
@@ -38,8 +37,7 @@ def parse_yaml(filepath: Union[str, Path]) -> Dict:
     if not filepath.exists():
         raise_error(f"File does not exist: {str(filepath.absolute())}")
     # Filepath reading
-    with open(filepath, "r") as f:
-        contents = yaml.safe_load(f)
+    contents = yaml.load(filepath)
```
LeSasse, line 40: Same Q as above

junifer/api/tests/test_functions.py
```diff
@@ -240,32 +247,37 @@ def test_queue_correct_yaml_config(
         The path to the test directory.
     monkeypatch : pytest.MonkeyPatch
         The monkeypatch object.
+    caplog : pytest.LogCaptureFixture
+        The logcapturefixture object.
```
LeSasse, line 251: Should you use CamelCase in the docstring here, similar to the DataGrabber issue?

junifer/api/tests/test_functions.py
```diff
@@ -240,32 +247,37 @@ def test_queue_correct_yaml_config(
         The path to the test directory.
     monkeypatch : pytest.MonkeyPatch
         The monkeypatch object.
```
LeSasse, line 249: Should you use CamelCase in the docstring here, similar to the DataGrabber issue?

junifer/api/utils.py
```diff
@@ -9,10 +9,19 @@
...
+
 from .._version import __version__
 from ..utils.logging import get_versions
 
 
+# Configure YAML class once for further use
+yaml = YAML()
+yaml.default_flow_style = False
+yaml.allow_unicode = True
+yaml.indent(mapping=2, sequence=4, offset=2)
```
LeSasse, line 22: So this is a global object that you import elsewhere to load files?

LeSasse (commented):
I just have a few clarification questions but that is it.

junifer/api/functions.py
synchon: That's cause `ruamel.yaml` allows you to write to files via `stream` argument in the `dump` function.

junifer/api/tests/test_functions.py
synchon: Yeah I was debating that and I followed what was there but fair point. I'll take care of that in a separate PR for all test functions.

junifer/api/utils.py
synchon: Yeah cause now you gotta have an instance of `YAML` class so doesn't make sense to create multiple objects if you have the same thing everywhere.

junifer/api/tests/test_functions.py
LeSasse: ok makes sense, fine by me!

junifer/api/functions.py
LeSasse: and this takes care of closing the ressource?

docs/faq.rst
synchon: As we have that constraint (`python=3.10`) in the `conda-env.yml` and also this syntax is incorrect from what I remember.

junifer/api/parser.py
synchon: Same argument as above.

junifer/api/tests/test_functions.py
synchon: Resolving this as I'll take care of this in a separate PR.

LeSasse (approved):
Approved, thanks for your hard work on this!

PR #236, opened 2023-06-28
[ENH]: Add `read()` method for storage-like objects

synchon:
* [x] tests added/passed

This PR adds a new `read()` method for `storage-like` objects. As of now, only `HDF5FeatureStorage` is implemented and allows to retrieve a feature as it is stored. This will allow users to do post-processing on the stored feature.

### DIFF
--- a/junifer/storage/sqlite.py
+++ b/junifer/storage/sqlite.py
@@ -258,7 +258,12 @@ class SQLiteFeatureStorage(PandasBaseFeatureStorage):
 
         """
         raise_error(
-            msg="read() is not implemented for SQLiteFeatureStorage.",
+            msg=(
+                "read() for SQLiteFeatureStorage is not currently planned to "
+                "be implemented in the near future. If you need the "
+                "functionality, contact the junifer developers "
+                "(https://juaml.github.io/junifer/main/help.html)."
+            ),
             klass=NotImplementedError,
         )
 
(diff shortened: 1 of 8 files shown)


junifer/storage/sqlite.py
```diff
@@ -235,6 +235,33 @@ def list_features(self) -> Dict[str, Dict[str, Any]]:
...
+            MD5 hash of the feature to read (default None).
+
+        Returns
+        -------
+        dict
+            The stored feature as a dictionary.
+
+        """
+        raise_error(
+            msg="read() is not implemented for SQLiteFeatureStorage.",
```
fraimondo, line 261: Can you rewrite this message so that users are hinted to contact the junifer devs if they need this feature? Something like that it is with the following: "This feature is not currently planned to be implemented in the near future. If you need such functionality, contact junifer developers (LINK TO HOW TO CONTACT)"

fraimondo (changes requested):

junifer/storage/sqlite.py
synchon: Maybe ask them to open an issue?

fraimondo: we have a link on the doc that can be used for that.

synchon: Are you referring to the Matrix channel link?

fraimondo: https://juaml.github.io/junifer/main/help.html

fraimondo (approved):

fraimondo:
docs are failing

synchon:
> docs are failing

That's coz of julearn example and it's failing to import from sklearn, my guess would be because of a new version of sklearn. So either we don't run the example for now or pin the version in julearn?

PR #245, opened 2023-08-10
[ENH]: Support for `Power 2013` coordinates

synchon:

We don't yet have the [Power 2013](https://www.jonathanpower.net/2013-neuron-hubs.html) coordinates in-built. This will be a great addition.

As other in-built coordinates.

_No response_

### Anything else to say?

_No response_


fraimondo (changes requested):
This is a backward's incompatible change. What about a deprecation cycle?

If I remember correctly, we can use functions to "load" coordinates/masks. Or we can implement it in a way that when someone uses "Power" it triggers a deprectation warning during a few release cycles. (We need to define).

synchon:
> This is a backward's incompatible change. What about a deprecation cycle?

That is true. I went with the breaking change as we are still not v1.0.0 . But, I don't mind putting a deprecation notice.

> If I remember correctly, we can use functions to "load" coordinates/masks. Or we can implement it in a way that when someone uses "Power" it triggers a deprectation warning during a few release cycles. (We need to define).

So, I'll put a deprecation notice stating that we'll remove it in v0.0.5 .

junifer/data/coordinates.py
```diff
@@ -134,9 +136,27 @@ def load_coordinates(name: str) -> Tuple[ArrayLike, List[str]]:
...
+    -----
+    DeprecationWarning
+        If ``Power`` is provided as the ``name``.
+
     """
     if name not in _available_coordinates:
         raise_error(f"Coordinates {name} not found.")
+
+    # Put up deprecation notice
+    if name == "Power":
```
fraimondo, line 150: I was thinking to use something like this: https://deprecation.readthedocs.io/en/latest/

fraimondo: while encapsulating the "power" in a function so we can keep track of the deprecation cycle. Otherwise we need to memorise stuff.

synchon: This would require us to include another dependency for junifer and for one release cycle I don't think it's worth it. I think this makes sense once we start having objects which require backwards incompatible changes and that would happen after v1.0.0 imo.

fraimondo: We'll create another PR for that. I don't want to have this messages all around. And this will happen quite often i'm afraid.

fraimondo (approved):

PR #182, opened 2023-08-24
[ENH]: Add support for subject(s) file for `junifer run`

synchon:

I think in most data processing projects it is desirable to be able to select specific subsets of all subjects. Sometimes one just wants to get all available data, but sometimes one may want only specific subsets (like for example only unrelated subjects or subjects matched on some other variable). In other cases one may want to just preprocess a specific subset for some initial exploration or testing before getting the full sample. 

Add a subject parameter to in-built data grabbers. Ideally in the full pipeline one can add a list of subjects to the yaml directly, or for example by providing a "subject.txt" file that lists all desired subjects. Something like that.

_No response_

### Anything else to say?

_No response_

### DIFF
--- a/junifer/api/cli.py
+++ b/junifer/api/cli.py
@@ -115,8 +115,12 @@ def _parse_elements_file(filepath: Path) -> List[Tuple[str, ...]]:
         index_col=False,  # no index column
         skipinitialspace=True,  # no leading space after delimiter
     )
+    # Remove trailing whitespace in cell entries
+    csv_df_trimmed = csv_df.apply(
+        lambda x: x.str.strip() if x.dtype == "object" else x
+    )
     # Convert to list of tuple of str
-    return list(map(tuple, csv_df.to_numpy().astype(str)))  # type: ignore
+    return list(map(tuple, csv_df_trimmed.to_numpy().astype(str)))
 
 
 def _validate_verbose(
--- a/junifer/api/cli.py
+++ b/junifer/api/cli.py
@@ -68,13 +68,10 @@ def _parse_elements(element: Tuple[str], config: Dict) -> Union[List, None]:
     # Check if the element is a file for single element;
     # if yes, then parse elements from it
     if len(element) == 1 and Path(element[0]).resolve().is_file():
-        fetched_element = _parse_elements_file(Path(element[0]).resolve())
+        elements = _parse_elements_file(Path(element[0]).resolve())
     else:
-        fetched_element = element
-    # Process multi-keyed elements
-    elements = [
-        tuple(x.split(",")) if "," in x else x for x in fetched_element
-    ]
+        # Process multi-keyed elements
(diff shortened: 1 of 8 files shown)


junifer/api/cli.py
```diff
@@ -32,27 +32,48 @@
...
-    # TODO: If len == 1, check if its a file, then parse elements from file
-    elements = [tuple(x.split(",")) if "," in x else x for x in element]
+    # Check if the element is a file for single element;
+    # if yes, then parse elements from it
+    if len(element) == 1 and Path(element[0]).resolve().is_file():
+        fetched_element = _parse_elements_file(Path(element[0]).resolve())
+    else:
+        fetched_element = element
+    # Process multi-keyed elements
+    elements = [
```
fraimondo, line 74: While this works, it's not quite robust. Splitting elements by `,` should be done only if there's no file involved.

junifer/api/cli.py
```diff
@@ -61,19 +82,41 @@ def _parse_elements(element: str, config: Dict) -> Union[List, None]:
...
-                "The 'elements' key is set in the configuration, but its value"
-                " is 'None'. It is likely that there is an empty 'elements' "
+                "The `elements` key is set in the configuration, but its value"
+                " is `None`. It is likely that there is an empty `elements` "
                 "section in the yaml configuration file."
             )
     return elements
 
 
+def _parse_elements_file(filepath: Path) -> List[str]:
```
fraimondo, line 99: This function should parse the file completely, giving the list of tuples required.

Why this?
Well, because we can have things like:
`sub-01,ses-01` or `sub-01, ses-01` or `sub-01, ses-01 `.

Using pandas to parse the CSV file would be easier.

fraimondo (changes requested):

junifer/api/cli.py
synchon: > Well, because we can have things like:
sub-01,ses-01 or sub-01, ses-01 or sub-01, ses-01 .

This worked but I anyway changed it to use `pandas` as you suggested.

junifer/api/cli.py
synchon: Addressed.

junifer/api/tests/test_cli.py
```diff
@@ -74,6 +83,59 @@ def test_run_and_collect_commands(
     assert collect_result.exit_code == 0
 
 
+@pytest.mark.parametrize(
```
fraimondo, line 86: This is not fully testing all the options.

What if the user wants to specify `subject` and `session`? Can you test that, it should be a comma-separated list.

fraimondo (changes requested):

junifer/api/tests/test_cli.py
synchon: Should be addressed now.

fraimondo (approved):

PR #252, opened 2023-09-27
[ENH]: Add support for native space

synchon:
* [x] tests added/passed

This PR provides support for converting to and operating on subject-native space.


junifer/data/coordinates.py
```diff
@@ -100,27 +173,128 @@ def register_coordinates(
...
+    """
+    # Load the coordinates
+    seeds, labels, _ = load_coordinates(name=coords)
+
+    # Transform coordinate if not in MNI space
+    if not target_data["space"].startswith("MNI"):
+        # Check for extra inputs
+        if extra_input is None:
+            raise_error(
+                "No extra input provided, requires `Warp` and `T1w` "
```
fraimondo, line 238: Can you be more explicit about what can't be done here?

junifer/data/masks.py
```diff
@@ -268,20 +332,82 @@ def get_mask(
             )
         mask_img = all_masks[0]
 
+    # Warp mask if not in MNI space
+    if not target_data["space"].startswith("MNI"):
+        # Check for extra inputs
+        if extra_input is None:
+            raise_error(
+                "No extra input provided, requires `Warp` and `T1w` "
```
fraimondo, line 340: Same here, what can't be done without the extra input provided?

junifer/data/parcellations.py
```diff
@@ -230,18 +382,33 @@ def load_parcellation(
...
             f"Parcellation {name} not found. "
             f"Valid options are: {list_parcellations()}"
         )
 
+    # Copy parcellation definition to avoid edits in original object
     parcellation_definition = _available_parcellations[name].copy()
     t_family = parcellation_definition.pop("family")
+    # Remove space conditionally
+    if t_family not in ["SUIT", "Tian"]:
+        space = parcellation_definition.pop("space")
```
fraimondo, line 407: Why this?

junifer/preprocess/fsl/apply_warper.py
```diff
@@ -0,0 +1,269 @@
...
+
+        Returns
+        -------
+        list of str
+            The list of data types that can be used as input for this
+            preprocessor.
+
+        """
+        # Constructed dynamically
+        return [self.on, self.ref, "Warp"]
```
fraimondo, line 83: This method should output the valid input that you can actually apply this transformer to. So in this case, is only BOLD (for the moment). Ref and Warp are requirements, but you can applywarp to the warp file.

fraimondo: Check here: https://github.com/juaml/junifer/blob/e3e942324b2f90f103a8500105e123d664139737/junifer/preprocess/confounds/fmriprep_confound_remover.py#L213

fraimondo: You need to override the `validate_input` method so it asks for all of the elements required.

junifer/data/coordinates.py
```diff
@@ -100,27 +173,128 @@ def register_coordinates(
...
+                "data types in particular."
+            )
+
+        # Create tempdir
+        tempdir = Path(tempfile.mkdtemp())
```
fraimondo, line 243: `mkdtemp` needs cleanup.

junifer/data/masks.py
```diff
@@ -268,20 +332,82 @@ def get_mask(
...
+                "data types in particular."
+            )
+
+        # Create tempdir
+        tempdir = Path(tempfile.mkdtemp())
```
fraimondo, line 345: `mkdtemp` needs cleanup.

junifer/data/parcellations.py
```diff
@@ -187,24 +201,162 @@ def list_parcellations() -> List[str]:
...
+    if not target_data["space"].startswith("MNI"):
+        # Check for extra inputs
+        if extra_input is None:
+            raise_error(
+                "No extra input provided, requires `Warp` and `T1w` "
+                "data types in particular."
+            )
+
+        # Create tempdir
+        tempdir = Path(tempfile.mkdtemp())
```
fraimondo, line 293: `mkdtemp` needs cleanup

fraimondo (changes requested):
Needs some more testing, but this is the most-likely definite codebase.

junifer/preprocess/fsl/apply_warper.py
synchon: Do you think it's a good way to do it considering one can do `super().__init__(on="BOLD")`. Overriding it defeats the point of the base method imo, but again I don't want to be pedantic about it.

junifer/data/parcellations.py
synchon: Because these two parcellation families already provide space as a parameter and if we pop the space then the retrieval functions won't work.

junifer/preprocess/fsl/apply_warper.py
synchon: This has been addressed as per our discussion to modify `BasePreprocessor`.

fraimondo (changes requested):

junifer/data/masks.py
fraimondo: Use the workdir manager now

junifer/data/parcellations.py
fraimondo: Workdir manager

junifer/preprocess/fsl/apply_warper.py
```diff
@@ -0,0 +1,269 @@
...
+        RuntimeError
+            If FSL commands fail.
+
+        """
+        # Get the min of the voxel sizes from input and use it as the
+        # resolution
+        resolution = np.min(input_data["data"].header.get_zooms()[:3])
+
+        # Create tempdir
+        tempdir = Path(tempfile.mkdtemp())
```
fraimondo, line 139: Workdir manager

fraimondo (changes requested):
Just needs to be updated after #254 is merged.

junifer/testing/datagrabbers.py
```diff
@@ -141,8 +144,8 @@ def get_item(self, subject: str) -> Dict[str, Dict]:
         anat_fname = self.datadir / f"{subject}_T1w.nii.gz"
         nib.save(fmri_img, fmri_fname)
         nib.save(anat_img, anat_fname)
-        out["BOLD"] = {"path": fmri_fname}
-        out["T1w"] = {"path": anat_fname}
+        out["BOLD"] = {"path": fmri_fname, "space": "MNI"}
+        out["T1w"] = {"path": anat_fname, "space": "native"}
```
fraimondo, line 148: Is this really in native space?

fraimondo (changes requested):

fraimondo (approved):

PR #254, opened 2023-10-04
[ENH]: Introduce `WorkDirManager`

synchon:
* [ ] tests added/passed

This PR introduces new `WorkDirManager` class which allows global access for temporary directories under the `workdir` parameter's values from the YAML, across the codebase.


junifer/pipeline/workdir_manager.py
```diff
@@ -0,0 +1,112 @@
...
+
+    Parameters
+    ----------
+    workdir : str or pathlib.Path, optional
+        The path to the super-directory. If None, "TMPDIR/junifer" is used
+        where TMPDIR is the platform-dependent temporary directory.
+
+    """
+
+    def __init__(self, workdir: Optional[Union[str, Path]] = None) -> None:
```
fraimondo, line 35: Since it's a singleton, any intialization should take care of cleaning up a previous one.

e.g:

```
WorkdirManager("/tmp/workdir1")
# Do some stuff that creates tempdirs
WorkdirManager("/tmp/workdir2")  # This should cleanup.
```

However, if a user does use it wrongly, it might cleanup in the middle.

So I would: 
1) add a `cleanup=False` parameter to the init
2) Raise an error if needs cleanup but cleanup is False

junifer/pipeline/workdir_manager.py
```diff
@@ -0,0 +1,112 @@
...
+    @property
+    def workdir(self) -> Path:
+        """Get working directory."""
+        return self._workdir
+
+    @workdir.setter
+    def workdir(self, value: Path) -> None:
+        """Set working directory."""
+        logger.debug(f"Changing working directory to {value}")
+        self._workdir = value
```
fraimondo, line 66: I would not allow to set the workdir like this, for the resasons stated in the constructor

junifer/pipeline/workdir_manager.py
```diff
@@ -0,0 +1,112 @@
...
+            The path to the temporary directory.
+
+        """
+        # Create root temporary directory if not created already
+        if self._root_temp_dir is None:
+            logger.debug(
+                f"Setting up temporary directory under {self._workdir}"
+            )
+            self._root_temp_dir = Path(
+                tempfile.mkdtemp(prefix=prefix, dir=self._workdir)
```
fraimondo, line 92: no prefix here, the root tmp dir should be random.

fraimondo (changes requested):

junifer/pipeline/workdir_manager.py
synchon: I've made the changes as you proposed. The problem is that the `__init__()` doesn't fire the second time you try to create an instance as it's a singleton, so the previous instance is returned.

fraimondo: Ok, so how can we fix it?

synchon: The only way I know of interacting with singletons in principle is to use getters and setters.

fraimondo: lets do that then

fraimondo (approved):

Issue #255, opened 2023-10-05
[DOC]: Document the precise spaces for each parcellation


fraimondo:
### Which element of the documentation do you want to modify and why?

For each existing parcellation, we should specify and document the precise space:

MNI152Lin
MNI152NLin2009aAsym
MNI152NLin2009aSym
MNI152NLin2009bAsym
MNI152NLin2009bSym
MNI152NLin2009cAsym
MNI152NLin2009cSym
MNI152NLin6Asym
MNI152NLin6Sym
MNI305
MNIColin27

PR #268, opened 2023-10-18
[ENH]: Add space awareness to existing components

synchon:
* [x] tests added/passed

This PR adds space awareness to existing components, enabling easy addition of spaces and transformation between them.

### DIFF
--- a/docs/builtin.rst
+++ b/docs/builtin.rst
@@ -247,6 +247,7 @@ Available
    * - Name
      - Options
      - Keys
+     - Spaces
      - Version added
      - Publication
    * - Schaefer
@@ -255,6 +256,7 @@ Available
        | ``Schaefer200x17``, ``Schaefer300x17``, ``Schaefer400x17``,
        | ``Schaefer500x17``, ``Schaefer600x17``, ``Schaefer700x17``,
        | ``Schaefer800x17``, ``Schaefer900x17``, ``Schaefer1000x17``
+     - ``MNI152NLin6Asym``
      - 0.0.1
      - | Schaefer, A., Kong, R., Gordon, E.M. et al.
        | Local-Global Parcellation of the Human Cerebral Cortex from
@@ -264,6 +266,7 @@ Available
    * - SUIT
      - ``space``
      - ``SUITxMNI``, ``SUITxSUIT``
+     - ``SUIT``, ``MNI152Lin6Asym``
      - 0.0.1
      - | Diedrichsen, J.
        | A spatially unbiased atlas template of the human cerebellum.
@@ -279,6 +282,7 @@ Available
        | ``TianxS2x3TxMNInonlinear2009cAsym``,
        | ``TianxS3x3TxMNInonlinear2009cAsym``,
        | ``TianxS4x3TxMNInonlinear2009cAsym``
+     - ``MNI152NLin6Asym``, ``MNI152NLin2009cAsym``
      - 0.0.1
      - | Tian, Y., Margulies, D.S., Breakspear, M. et al.
        | Topographic organization of the human subcortex
@@ -288,6 +292,7 @@ Available
    * - AICHA
      - ``version``
      - ``AICHA_v1``, ``AICHA_v2``
+     - ``MNI152Lin6Asym``
      - 0.0.3
      - | Joliot, M., Jobard, G., Naveau, M. et al.
        | AICHA: An atlas of intrinsic connectivity of homotopic areas.
@@ -297,6 +302,7 @@ Available
      - ``year``, ``n_rois``
      - | ``Shen_2013_50``, ``Shen_2013_100``, ``Shen_2013_150``,
        | ``Shen_2015_268``, ``Shen_2019_368``
+     - ``MNI152NLin2009cAsym``
      - 0.0.3
      - | Shen, X., Tokoglu, F., Papademetris, X., Constable, R.T.
        | Groupwise whole-brain parcellation from resting-state fMRI data
@@ -322,6 +328,7 @@ Available
        | ``Yan400xKong17``, ``Yan500xKong17``, ``Yan600xKong17``,
        | ``Yan700xKong17``, ``Yan800xKong17``, ``Yan900xKong17``,
        | ``Yan1000xKong17``
+     - ``MNI152NLin6Asym``
      - 0.0.3
      - | Yan, X., Kong, R., Xue, A., et al.
        | Homotopic local-global parcellation of the human cerebral cortex from
@@ -590,28 +597,33 @@ Available
 
    * - Name
      - Keys
+     - Spaces
      - Version added
      - Description - Publication
    * - Vickery-Patil (Gray Matter)
      - | ``GM_prob0.2``
+     - ``MNI152Lin6Asym``
      - 0.0.1
      - | Vickery, Sam, & Patil, Kaustubh. (2022).
        | Chimpanzee and Human Gray Matter Masks [Data set]. Zenodo.
        | https://doi.org/10.5281/zenodo.6463123
    * - Vickery-Patil (Cortex + Basal Ganglia)
      - | ``GM_prob0.2_cortex``
+     - ``MNI152Lin6Asym``
      - 0.0.1
      - | Vickery, Sam, & Patil, Kaustubh. (2022).
        | Chimpanzee and Human Gray Matter Masks [Data set]. Zenodo.
        | https://doi.org/10.5281/zenodo.6463123
    * - Nilearn's MNI152 1mm-resolution mask
      - | ``compute_brain_mask``
+     - Adapts to the target data
      - 0.0.2
      - | Compute the whole-brain mask. This mask is calculated using
        | MNI152 1mm-resolution template mask onto the target image.
        | See :func:`nilearn.masking.compute_brain_mask`
    * - Nilearn's mask computed from FMRI data
      - | ``compute_epi_mask``
+     - Adapts to the target data
      - 0.0.2
      - | Compute a brain mask from fMRI data. This is based on an heuristic
        | proposed by T.Nichols: find the least dense point of the histogram,
@@ -619,6 +631,7 @@ Available
        | image histogram. See :func:`nilearn.masking.compute_epi_mask`
    * - Nilearn's background mask
      - | ``compute_background_mask``
+     - Adapts to the target data
      - 0.0.2
      - | Compute a brain mask for the images by guessing the value of the
        | background from the border of the image.
@@ -626,6 +639,7 @@ Available
 
    * - Nilearn's ICBM152 template gray-matter mask
      - | ``fetch_icbm152_brain_gm_mask``
+     - ``MNI152NLin2009aAsym``
      - 0.0.2
      - | Compute a gray-matter mask from the asymmetrical ICBM152 2009 template,
        | release a.
--- a/examples/run_compute_parcel_mean.py
+++ b/examples/run_compute_parcel_mean.py
@@ -2,18 +2,21 @@
 Computer Parcel Aggregation.
 ============================
 
-This example uses a ParcelAggregation marker to compute the mean of each parcel
-using the Schaefer parcellations (100 rois, 7 Yeo networks) for both a 3D and
-4D nifti
+This example uses the ``ParcelAggregation`` marker to compute the mean of each
+parcel using the Schaefer parcellations (100 rois, 7 Yeo networks) for both 3D
+and 4D NIfTI.
 
-Authors: Federico Raimondo
+Authors: Federico Raimondo, Synchon Mandal
 
 License: BSD 3 clause
 """
 
-import nilearn
-
-from junifer.markers.parcel_aggregation import ParcelAggregation
+from junifer.testing.datagrabbers import (
+    OasisVBMTestingDataGrabber,
+    SPMAuditoryTestingDataGrabber,
+)
+from junifer.datareader import DefaultDataReader
+from junifer.markers import ParcelAggregation
 from junifer.utils import configure_logging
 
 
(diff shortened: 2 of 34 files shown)


synchon:
> Does it make sense to tackle #255 on this one to? Basically instead of using `MNI` use the right MNI and then raise error if the space does not match. Otherwise this will be corrected in the next days, giving an somehow unstable main branch.

Yes totally. That's my plan as well. Will spend some time to get it correct and then we can add native space with a couple of commits.

fraimondo:
> > Does it make sense to tackle #255 on this one to? Basically instead of using `MNI` use the right MNI and then raise error if the space does not match. Otherwise this will be corrected in the next days, giving an somehow unstable main branch.
> 
> Yes totally. That's my plan as well. Will spend some time to get it correct and then we can add native space with a couple of commits.

perfect, will wait to review then.

synchon:
> > > > > Does it make sense to tackle #255 on this one to? Basically instead of using `MNI` use the right MNI and then raise error if the space does not match. Otherwise this will be corrected in the next days, giving an somehow unstable main branch.
> > > > 
> > > > 
> > > > Yes totally. That's my plan as well. Will spend some time to get it correct and then we can add native space with a couple of commits.
> > > 
> > > 
> > > perfect, will wait to review then.
> > 
> > 
> > Should be good to review now.
> 
> Ok my bad, I corrected some stuff which broke the tests, will fix it.

Good for review now.

docs/builtin.rst
```diff
@@ -279,6 +282,7 @@ Available
        | ``TianxS2x3TxMNInonlinear2009cAsym``,
        | ``TianxS3x3TxMNInonlinear2009cAsym``,
        | ``TianxS4x3TxMNInonlinear2009cAsym``
+     - ``MNI152NLin6Asym``, ``MNI152NLin2009cAsym``
```
fraimondo, line 285: Since we add the `space` parameter, we need to remove the space from the name.

docs/builtin.rst
```diff
@@ -264,6 +266,7 @@ Available
    * - SUIT
      - ``space``
      - ``SUITxMNI``, ``SUITxSUIT``
```
fraimondo, line 268: We need to remove the space from the name.

docs/builtin.rst
```diff
@@ -590,42 +597,49 @@ Available
...
    * - Vickery-Patil (Gray Matter)
      - | ``GM_prob0.2``
+     - ``MNI152Lin6Asym``
      - 0.0.1
      - | Vickery, Sam, & Patil, Kaustubh. (2022).
        | Chimpanzee and Human Gray Matter Masks [Data set]. Zenodo.
        | https://doi.org/10.5281/zenodo.6463123
    * - Vickery-Patil (Cortex + Basal Ganglia)
      - | ``GM_prob0.2_cortex``
+     - ``MNI152Lin6Asym``
```
fraimondo, line 612: How sure are we that is the Linear 6th Gen Asymmetric?

junifer/data/parcellations.py
```diff
@@ -42,7 +45,7 @@
 # TODO: have separate dictionary for built-in
 _available_parcellations: Dict[str, Dict[Any, Any]] = {
     "SUITxSUIT": {"family": "SUIT", "space": "SUIT"},
-    "SUITxMNI": {"family": "SUIT", "space": "MNI"},
+    "SUITxMNI": {"family": "SUIT", "space": "MNI152Lin6Asym"},
```
fraimondo, line 48: Name should be SUIT

junifer/data/masks.py
```diff
@@ -219,46 +256,76 @@ def get_mask(
...
                     copy=True,
                 )
+            all_spaces.append(mask_space)
         all_masks.append(mask_img)
+
+    # Multiple masks, need intersection / union
     if len(all_masks) > 1:
-        mask_img = intersect_masks(all_masks, **intersect_params)
+        # Intersect / union of masks only if all masks are in the same space
+        if len(set(all_spaces)) == 1 or all(
```
fraimondo, line 313: We can't intersect masks in different MNI spaces.

junifer/data/parcellations.py
```diff
@@ -61,27 +65,28 @@
         "family": "Tian",
         "scale": scale,
         "magneticfield": "7T",
-        "space": "MNI6thgeneration",
+        "space": "MNI152NLin6Asym",
```
fraimondo, line 68: Same here for the names, it should not contain the space now that we have a dedicated field.

junifer/data/parcellations.py
```diff
@@ -236,17 +258,91 @@ def get_parcellation(
...
         labels = all_labels[0]
     else:
-        # Merge the parcellations
-        resampled_parcellation_img, labels = merge_parcellations(
-            parcellations_list=all_parcellations,
-            parcellations_names=parcellation,
-            labels_lists=all_labels,
+        # Merge the parcellations only if all parcellations are in the same
+        # space
+        if len(set(all_spaces)) == 1 or all(
```
fraimondo, line 270: we only merge when it's the same space.

junifer/datagrabber/aomic/id1000.py
```diff
@@ -121,6 +121,10 @@ def get_item(self, subject: str) -> Dict:
         out = super().get_item(subject=subject)
         if out.get("BOLD"):
             out["BOLD"]["mask_item"] = "BOLD_mask"
+            # Add space information
+            out["BOLD"].update({"space": "MNI"})
```
fraimondo, line 125: We need to find the right MNI space version.

junifer/datagrabber/aomic/piop2.py
```diff
@@ -166,6 +166,10 @@ def get_item(self, subject: str, task: str) -> Dict:
         out = super().get_item(subject=subject, task=f"{task}_acq-seq")
         if out.get("BOLD"):
             out["BOLD"]["mask_item"] = "BOLD_mask"
+            # Add space information
+            out["BOLD"].update({"space": "MNI"})
```
fraimondo, line 170: same here

junifer/datagrabber/aomic/piop1.py
```diff
@@ -166,8 +166,12 @@ def get_item(self, subject: str, task: str) -> Dict:
         out = super().get_item(subject=subject, task=new_task)
         if out.get("BOLD"):
             out["BOLD"]["mask_item"] = "BOLD_mask"
+            # Add space information
+            out["BOLD"].update({"space": "MNI"})
```
fraimondo, line 170: get right space name

junifer/testing/datagrabbers.py
```diff
@@ -55,7 +55,10 @@ def get_item(self, subject: str) -> Dict[str, Dict]:
         """
         out = {}
         i_sub = int(subject.split("-")[1]) - 1
-        out["VBM_GM"] = {"path": Path(self._dataset.gray_matter_maps[i_sub])}
+        out["VBM_GM"] = {
+            "path": Path(self._dataset.gray_matter_maps[i_sub]),
+            "space": "MNI",
```
fraimondo, line 60: I'd rather use the right space here, so we test correctly the masks/parecellations.

junifer/testing/datagrabbers.py
```diff
@@ -141,8 +144,8 @@ def get_item(self, subject: str) -> Dict[str, Dict]:
         anat_fname = self.datadir / f"{subject}_T1w.nii.gz"
         nib.save(fmri_img, fmri_fname)
         nib.save(anat_img, anat_fname)
-        out["BOLD"] = {"path": fmri_fname}
-        out["T1w"] = {"path": anat_fname}
+        out["BOLD"] = {"path": fmri_fname, "space": "MNI"}
```
fraimondo, line 147: I'd rather use the right space here, so we test correctly the masks/parecellations.

junifer/testing/datagrabbers.py
```diff
@@ -236,12 +239,14 @@ def get_item(self, subject: str) -> Dict[str, Dict]:
         """
         out = {}
         i_sub = int(subject.split("-")[1]) - 1
-        out["BOLD"] = {"path": Path(self._dataset["func"][i_sub])}
-        conf_format = "fmriprep"
-
+        out["BOLD"] = {
+            "path": Path(self._dataset["func"][i_sub]),
+            "space": "MNI",
```
fraimondo, line 244: I'd rather use the right space here, so we test correctly the masks/parecellations.

junifer/testing/datagrabbers.py
```diff
@@ -236,12 +239,14 @@ def get_item(self, subject: str) -> Dict[str, Dict]:
...
+        }
         out["BOLD_confounds"] = {
             "path": Path(self._dataset["confounds"][i_sub]),
-            "format": conf_format,
+            "format": "fmriprep",
+            "space": "MNI",
```
fraimondo, line 249: Do we need "space" for the confounds file? It does not make any sense to me.

fraimondo (changes requested):

docs/builtin.rst
synchon: The mask was made with CAT12 which uses SPM12 and that uses its own template which is similar to MNI Linear 6th Generation Asymmetric, at least that's what I found out given the not-so-proper documentation.

examples/run_compute_parcel_mean.py
fraimondo, line 1: Is this example running? I mean, only _coordinates_ are in "MNI" space. The rest should be in the specific MNI space.

docs/builtin.rst
fraimondo: Can you check with @kaurao and Sam? They might have change the template. Mainly what I'm doubting is the Linear/NLinear difference.

junifer/data/masks.py
```diff
@@ -219,46 +256,74 @@ def get_mask(
...
+
+    # Multiple masks, need intersection / union
     if len(all_masks) > 1:
-        mask_img = intersect_masks(all_masks, **intersect_params)
+        # Intersect / union of masks only if all masks are in the same space
+        if len(set(all_spaces)) == 1:
+            mask_img = intersect_masks(all_masks, **intersect_params)
+        else:
+            # Check for inherited masks with target image in MNI space
+            if "inherit" in all_spaces and target_data["space"].startswith(
```
fraimondo, line 317: I think at this point we should not have the "inherit" as it was already computed before. Usually the "inherit" is a callable.

junifer/data/masks.py
```diff
@@ -268,20 +333,82 @@ def get_mask(
             )
         mask_img = all_masks[0]
 
+    # Warp mask if not in MNI space
+    if not target_data["space"].startswith("MNI"):
```
fraimondo, line 337: This will have conflicts with other non-MNI spaces. I think we need to check for either "native" or not any standard space supported by junifer.

junifer/data/parcellations.py
fraimondo: To be done in another PR

junifer/data/parcellations.py
fraimondo: To be done in another PR

junifer/data/parcellations.py
```diff
@@ -61,27 +65,28 @@
...
         "magneticfield": "3T",
-        "space": "MNInonlinear2009cAsym",
+        "space": "MNI152NLin2009cAsym",
     }
 # Add AICHA parcellation info
 for version in (1, 2):
     _available_parcellations[f"AICHA_v{version}"] = {
         "family": "AICHA",
         "version": version,
+        "space": "MNI152Lin6Asym",
```
fraimondo, line 89: We need confirmation that this is the exact space.

junifer/data/parcellations.py
```diff
@@ -236,17 +258,89 @@ def get_parcellation(
...
+                labels_lists=all_labels,
+            )
+        else:
+            raise_error(
+                msg="Parcellations are in different spaces, unable to merge.",
+                klass=RuntimeError,
+            )
+
+    # Warp parcellation if not in MNI space
+    if not target_data["space"].startswith("MNI"):
```
fraimondo, line 283: Same as with masks. MNI is only one of the spaces we support. We also have SUIT. This will fail for non-native and non-MNI space.

junifer/data/parcellations.py
```diff
@@ -805,6 +908,9 @@ def _retrieve_suit(
     resolution = closest_resolution(resolution, _valid_resolutions)
 
     # define file names
+    # Format space if MNI; required for the file name
+    if space.startswith("MNI"):
```
fraimondo, line 912: I would leave the right space and not just MNI in the filename. Even if this triggers a second download on old users.

fraimondo (changes requested):
We need to update the tests so the data is not in "MNI" space but in "MNINLin6..." or something specific.

junifer/data/masks.py
synchon: I wanted to get the native check in the native support PR. About standard spaces supported by junifer, there's not really a list of them now but in that case we need to make one, would you agree to making one?

junifer/data/parcellations.py
synchon: To be exactly specific, this is what SPM 12 uses: `IXI549Space` ([ref](https://bids-specification.readthedocs.io/en/latest/appendices/coordinate-systems.html)). Now, on the MNI side, it's close to `MNI152Lin6Asym` ([ref](https://www.lead-dbs.org/about-the-mni-spaces/)). So, I can change it to `IXI549Space` if you agree.

docs/builtin.rst
synchon: I checked and it's based on the TPM and Shooting GM masks from CAT12 which uses SPM12. SPM12 uses `IXI549Space` as its space and on the MNI side it's closest to `MNI152Lin6Asym`.

junifer/data/masks.py
fraimondo: For me it does not make much sense to have a "list" of supported at this moment. A user can create a dataset with space "XYZ" and then add a mask/parcellation for such space and it should work.

Here the condition to test is "native" and not if it's MNI. As the idea is to warp from whatever the mask is to "native" space. Now this "whatever" needs to match the available transform space.

junifer/data/parcellations.py
fraimondo: Is not what I meant. They provide the atlas for SPM12, and that table shows the default. 

From the paper:

Pre-processing was performed using [Statistical Parametric Mapping](https://www.sciencedirect.com/topics/neuroscience/statistical-parametric-mapping) subroutines (SPM5)

Later on:

normalized to the stereotaxic space of the Montreal Neurological Institute (MNI) template ([Ashburner and Friston, 2005]

But they don't specify which MNI space.


From the README of v2:

- "AICHA was projected in the MNI space as defined by the SPM12 template ". I guess this is the IXI549Space


But again, we don't know much about the v1 or the "guess".

synchon: When you check https://www.gin.cnrs.fr/en/tools/aicha/ you would see they offer both v1 and v2 for SPM12 which we use. Now after that, it goes back to the reply I had earlier.

fraimondo: New take:

1) Set it to `IXI549Space`
2) Raise a "warning" when this atlas is being used as we are not so sure about the space.
3) Contact the authors to clarify.

What do you think?

synchon: Sounds good.

fraimondo (approved):

PR #272, opened 2023-10-31
[FIX]: Fix `_check_fsl()`

synchon:
* [ ] tests added/passed

This PR fixes `_check_fsl()` by replacing 0 with 1 as the return code for successful FSL commands.


junifer/pipeline/utils.py
```diff
@@ -135,7 +135,7 @@ def _check_fsl(commands: Optional[List[str]] = None) -> bool:
         shell=True,  # is unsafe but kept for resolution via PATH
         check=False,
     )
-    fsl_found = completed_process.returncode == 0
+    fsl_found = completed_process.returncode == 1
```
fraimondo, line 138: the issue here is that `flirt -v` is not the right test.

```
(miniconda3)juseless ➜  ~  flirt -v
Singularity args: 
Corrected args for FSL: flirt -v
Running command: /data/group/appliedml/tools/fsl_6.0.4-patched2/singularity_cmd exec  /data/group/appliedml/tools/fsl_6.0.4-patched2/fsl_6.0.4-patched2.sif flirt -v
ERROR:: Input volume filename not found

FLIRT version 6.0

Usage: /usr/local/fsl/bin/flirt [options] -in <inputvol> -ref <refvol> -out <outputvol>
       /usr/local/fsl/bin/flirt [options] -in <inputvol> -ref <refvol> -omat <outputmatrix>
       /usr/local/fsl/bin/flirt [options] -in <inputvol> -ref <refvol> -applyxfm -init <matrix> -out <outputvol>
       
```
Then it has a lot on the usage of flirt. And this is the rest:

```
(miniconda3)juseless ➜  ~  echo $?
2
(miniconda3)juseless ➜  ~  flirt -version
Singularity args: 
Corrected args for FSL: flirt -version
Running command: /data/group/appliedml/tools/fsl_6.0.4-patched2/singularity_cmd exec  /data/group/appliedml/tools/fsl_6.0.4-patched2/fsl_6.0.4-patched2.sif flirt -version
FLIRT version 6.0
(miniconda3)juseless ➜  ~  echo $?
0

```

fraimondo (changes requested):

junifer/pipeline/utils.py
synchon: > the issue here is that flirt -v is not the right test.

Yes my bad, I forgot to change that, it should be `flirt -version`. The issue here is that unlike `afni` from which you can call `3dReHo` (for example) and you get back status of 0, FSL commands which we use like `applywarp`, don't have versions and if you call the command as is you would get back status of 1 and that is true even if you call for help:

```
❱ applywarp
Singularity args:
Corrected args for FSL: applywarp
Running command: /data/group/appliedml/tools/fsl_6.0.4-patched2/singularity_cmd exec  /data/group/appliedml/tools/fsl_6.0.4-patched2/fsl_6.0.4-patched2.
sif applywarp
***************************************************
The following COMPULSORY options have not been set:
        -i,--in filename of input image (to be warped)
        -r,--ref        filename for reference image
        -o,--out        filename for output (warped) image
***************************************************

Part of FSL (ID: 6.0.4:ddd0a010)
applywarp
Copyright(c) 2001, University of Oxford (Jesper Andersson)
.
.
.
        -v,--verbose    switch on diagnostic messages
        -h,--help       display this message
```
```
❱ echo $?                                                                                                                                          
1
```
So, for our use-cases, `flirt -version` would be an outlier which returns 0, so we could change the hardcoded check to 0 and make everything else check for 1.

fraimondo: You are right. So I see two options:

1) We can test that is not 127 (command not found)
2) We leave a comment saying that this works, so far, as long as the tested command in FSL returns 1 when executed with `-h` (help, that I think is in every command)

What do you think?

fraimondo: Mainly because of this:

```
(miniconda3)juseless ➜  ~  fslfft -v
Singularity args: 
Corrected args for FSL: fslfft -v
Running command: /data/group/appliedml/tools/fsl_6.0.4-patched2/singularity_cmd exec  /data/group/appliedml/tools/fsl_6.0.4-patched2/fsl_6.0.4-patched2.sif fslfft -v
Usage: /usr/local/fsl/bin/fslfft <input volume> <output volume> [-inv]

e.g.   /usr/local/fsl/bin/fslfft invol outvol
(miniconda3)juseless ➜  ~  echo $?
255
```

synchon: Agreed, let's go with option 1, will make the changes.

fraimondo (approved):

PR #282, opened 2023-12-09
[BUG]: Propagate ReHo and fALFF maps' file paths

synchon:
* [x] tests added/passed

This PR enables file path propagation for ReHo and fALFF maps from the estimator to the compute step so that aggregation markers can use it for conversion to other template spaces if and when required.

### DIFF
--- a/junifer/markers/falff/tests/test_falff_estimator.py
+++ b/junifer/markers/falff/tests/test_falff_estimator.py
@@ -55,7 +55,7 @@ def test_ALFFEstimator_cache_python(tmp_path: Path) -> None:
 
     # Compute again with cache, should be faster
     start_time = time.time()
-    alff, falff, alff_path, falff_path = estimator.fit_transform(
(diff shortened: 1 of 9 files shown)


fraimondo (approved):

junifer/markers/falff/tests/test_falff_estimator.py
```diff
@@ -149,7 +151,7 @@ def test_ALFFEstimator_cache_afni(tmp_path: Path) -> None:
 
     # Compute again with cache, should be faster
     start_time = time.time()
-    alff, falff = estimator.fit_transform(
+    alff, falff, alff_path, falff_path = estimator.fit_transform(
         use_afni=True,
```
fraimondo, line 155: Needs to be updated

fraimondo (changes requested):
Can't put the comment on the right line, but the test FALFF with afni needs to be udpated, it fails on the first `estimator.fit_transform`

PR #286, opened 2023-12-15
[BUG]: Correct path propagation logic for `ReHoEstimator` and `ALFFEstimator`

synchon:
* [x] tests added/passed

This PR fixes the path propagation logic for `ReHoEstimator` and `ALFFEstimator` based on the input data space. If the input data space is "native", the original input data path is passed down for further use in `get_coordinates()` (as of now) for transforming coordinates to subject-"native" template space, else the actual ReHo and (f)ALFF map paths are passed down.

### DIFF
--- a/docs/changes/newsfragments/286.bugfix
+++ b/docs/changes/newsfragments/286.bugfix
@@ -1,0 +1,1 @@
+Pass down input path if input space is "native" for ``ReHoEstimator`` and ``ALFFEstimator``, else use respective compute maps by `Fede Raimondo`_ and `Synchon Mandal`_
--- a/junifer/markers/falff/falff_base.py
+++ b/junifer/markers/falff/falff_base.py
@@ -148,6 +148,9 @@ class ALFFBase(BaseMarker):
 
         estimator = ALFFEstimator()
 
+        # If the input data space is "native", then alff_path and falff_path
+        # both point to the input data path as it might be required to use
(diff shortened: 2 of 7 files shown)


fraimondo:
I don't fully agree with this solution though. I still do believe that the data and the path should point to the same object.

Nevertheless, this is a patch for the current main branch.

I would like to see this fixed when we have multi-mni space support. In that case, the img2imgcoord could simply get the T1w of the right space from templateflow and then we will not need this patches.

Can you please create an issue so we rollback this hack in the future?

synchon:
> I don't fully agree with this solution though. I still do believe that the data and the path should point to the same object.
> 
> Nevertheless, this is a patch for the current main branch.
> 
> I would like to see this fixed when we have multi-mni space support. In that case, the img2imgcoord could simply get the T1w of the right space from templateflow and then we will not need this patches.
> 
> Can you please create an issue so we rollback this hack in the future?

I don't have a better solution atm, please do share if you have a better one. I agree with you that the data and the path should point to the same object. The next thing we want is multi-MNI template space support, so if the solution is not too convincing, we can close this PR and get the multi-MNI support which would fix this as well. Of course, if you think we go with this, I'll make an issue to revert this.

fraimondo:
For the moment this works and it's ok. I just want to document that this needs to be removed when multi-MNI space is added.

fraimondo (approved):

PR #293, opened 2024-01-10
[ENH]: Introduce `_AntsApplyTransformsWarper`

synchon:
* [ ] tests added/passed

This PR introduces `_AntsApplyTransformsWarper` as a generic ANTs `antsApplyTransforms` wrapper class for internal use. It also adapts `BOLDWarper`, `get_mask`, `get_parcellation` and `get_coordinates` to dynamically choose FSL or ANTs depending on warp file extension. It also completes unit tests for `_ApplyWarper`.


junifer/preprocess/bold_warper.py
```diff
@@ -35,8 +36,13 @@ class BOLDWarper(BasePreprocessor):
     ] = [
         {
             "name": "fsl",
-            "optional": False,
-            "commands": ["applywarp"],
+            "optional": True,
```
fraimondo, line 39: How does this work with the dependency check?

synchon: In the `BOLDWarper`, the check is optional, but when the actual preprocessor class i.e., `ApplyWarper` and `AntsApplyTransformsWarper` are initialised inside it, the checks are mandatory and fails if the tools are not found.

fraimondo: Excellent!

fraimondo (approved):

PR #297, opened 2024-01-30
[ENH]: Introduce `get_xfm` for fetching transformation files

synchon:
* [x] tests added/passed

This PR introduces a new function `junifer.data.get_xfm()` to fetch transformation files for moving between template spaces.

### DIFF
--- a/junifer/data/template_spaces.py
+++ b/junifer/data/template_spaces.py
@@ -13,7 +13,7 @@ from ..utils import logger, raise_error
 
 def get_xfm(
     src: str, dst: str, xfms_dir: Union[str, Path, None] = None
-) -> Path:
+) -> Path:  # pragma: no cover
     """Fetch warp files to convert from ``src`` to ``dst``.
 
     Parameters
--- a/junifer/data/template_spaces.py
+++ b/junifer/data/template_spaces.py
@@ -1,0 +1,91 @@
+"""Provide functions for template spaces."""
+
+# Authors: Synchon Mandal <s.mandal@fz-juelich.de>
+# License: AGPL
+
+from pathlib import Path
+from typing import Union
+
+import httpx
+
+from ..utils import logger, raise_error
+
+
+def get_xfm(
+    src: str, dst: str, xfms_dir: Union[str, Path, None] = None
+) -> Path:
+    """Fetch warp files to convert from ``src`` to ``dst``.
+
+    Parameters
+    ----------
+    src : str
+        The template space to transform from.
+    dst : str
+        The template space to transform to.
+    xfms_dir : str or pathlib.Path, optional
+        Path where the retrieved transformation files are stored.
+        The default location is "$HOME/junifer/data/xfms" (default None).
+
+    Returns
+    -------
+    pathlib.Path
+        The path to the transformation file.
+
+    Raises
+    ------
+    RuntimeError
+        If there is a problem fetching files.
+
+    """
+    if xfms_dir is None:
+        xfms_dir = Path().home() / "junifer" / "data" / "xfms"
+        logger.debug(f"Creating xfm directory at: {xfms_dir.resolve()}")
+        # Create default junifer data directory if not present
+        xfms_dir.mkdir(exist_ok=True, parents=True)
+    # Convert str to Path
+    elif not isinstance(xfms_dir, Path):
+        xfms_dir = Path(xfms_dir)
+
+    # Set local file prefix
+    xfm_file_prefix = f"{src}_to_{dst}"
+    # Set local file dir
+    xfm_file_dir = xfms_dir / xfm_file_prefix
+    # Create local directory if not present
+    xfm_file_dir.mkdir(exist_ok=True, parents=True)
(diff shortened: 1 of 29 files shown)


junifer/data/template_spaces.py
```diff
@@ -0,0 +1,91 @@
...
+    if xfm_file_path.exists():
+        logger.info(
+            f"Found existing xfm file for {src} to {dst} at "
+            f"{xfm_file_path.resolve()}"
+        )
+        return xfm_file_path
+
+    # Set URL
+    url = (
+        "https://gin.g-node.org/synchon/human-template-xfms/raw/main/xfms/"
```
fraimondo, line 69: Can we make it the juaml organization in g-node.org?

junifer/data/template_spaces.py
```diff
@@ -0,0 +1,91 @@
...
+            resp.raise_for_status()
+        except httpx.HTTPError as exc:
+            raise_error(
+                f"Error response {exc.response.status_code} while "
+                f"requesting {exc.request.url!r}",
+                klass=RuntimeError,
+            )
+        else:
+            with open(xfm_file_path, "ab") as f:
+                for chunk in resp.iter_bytes():
```
fraimondo, line 88: Is `iter_bytes` still streaming? or the content is already in memory? If it is still streaming, an exception might occur here an it will not be catch.

junifer/data/tests/test_template_spaces.py
```diff
@@ -0,0 +1,30 @@
...
+import socket
+from pathlib import Path
+
+import pytest
+
+from junifer.data import get_xfm
+
+
+@pytest.mark.skipif(
+    socket.gethostname() != "juseless",
```
fraimondo, line 15: Why only for juseless?

pyproject.toml
```diff
@@ -41,11 +41,12 @@ dependencies = [
     "datalad>=0.15.4,<0.20",
     "pandas>=1.4.0,<2.2",
     "nibabel>=3.2.0,<5.11",
-    "nilearn>=0.9.0,<=0.11.0",
+    "nilearn>=0.9.0,<=0.10.2",
     "sqlalchemy>=1.4.27,<=2.1.0",
     "ruamel.yaml>=0.17,<0.18",
     "importlib_metadata; python_version<'3.10'",
     "h5py>=3.8.0,<3.10",
+    "httpx[http2]==0.26.0",
```
fraimondo, line 49: Is it that strict the dependency?

junifer/data/template_spaces.py
synchon: Yeah we can, since I don't have permission to make a repo in the org, I made it in my account.

junifer/data/template_spaces.py
synchon: It's a generator basically thus it's still streaming. I don't find any exceptions being raised there in `httpx`'s source code.

junifer/data/tests/test_template_spaces.py
synchon: The download takes well around 10 mins.

pyproject.toml
synchon: It's always recommended to keep it like so when dealing with libraries which aren't 1.0 yet as the APIs can change very often.

junifer/data/template_spaces.py
fraimondo: Now you have !

junifer/data/template_spaces.py
fraimondo: Not from httpx, but mostly from an underlying connection issue if the connection is cut

junifer/data/tests/test_template_spaces.py
fraimondo: damn, we need to create a "slow test" suite that we use it to check only when about to merge or something like that. For the moment we keep it like this.

junifer/data/template_spaces.py
synchon: You will get error from `httpcore`, which is the underlying HTTP library, like you get from `urllib3` when using `requests`.

junifer/data/tests/test_template_spaces.py
synchon: Yeah that's a good idea. Well the essential issue is that Gin is not really as powerful as GitHub when you consider infrastructure, so the octet-stream downloads are not cached from what I understand.

junifer/data/template_spaces.py
synchon: Transferred.

junifer/data/template_spaces.py
fraimondo: Exactly, I thought of catching any exception and giving a more user-friendly error message of what's going

synchon: Ideally it should be dealt by the underlying http lib. You won't get an exception to catch when doing `iter_bytes()`, having an exception block there would not server the purpose imo.

fraimondo (approved):

PR #298, opened 2024-01-31
[ENH]: Introduce `get_template` for getting templates

synchon:
* [x] tests added/passed

This PR introduces a new function `junifer.data.get_template()` to get template space images via `templateflow`, tailored to a target data.


synchon:
Should be merged after #297

junifer/data/template_spaces.py
```diff
@@ -89,3 +92,63 @@ def get_xfm(
...
+        raise_error(f"Unknown template space: {space}")
+
+    # Get the min of the voxels sizes and use it as the resolution
+    target_img = target_data["data"]
+    resolution = np.min(target_img.header.get_zooms()[:3]).astype(int)
+
+    logger.info(f"Downloading template {space} in resolution {resolution}")
+    # Retrieve template
+    try:
+        template_path = tflow.get(
```
fraimondo, line 139: We need to get the "closest resolution" here

fraimondo (changes requested):

junifer/data/template_spaces.py
synchon: The `closest_resolution` function takes a list of valid resolutions, but we don't have that info right away. For that we need an extra scraping of the templateflow metadata which adds a bit of overhead. I don't get the complete rationale, maybe I can implement it in a different way if I understand correctly.

fraimondo: The idea of that function is to tell you which is the resolution of the parcellation/mask/etc that you need to use for the image you have.

synchon: We are already getting the resolution for the template required from the target data and it needs to be int for templateflow. If templateflow doesn't have that, it throws an error. So in that case, we need to resample.

synchon: But in our case, we have it and we fetch accordingly.

fraimondo: no, we don't have "any" resolution. We have some resolutions. Templateflow is the same. We have A, B, C and D as resolutions. Let's say our image is in resolution E. We need to get the "closest" to E following the criteria defined in the "closest resolution" function.

synchon: So there are two options:

- scrape templateflow metadata to get updated info but has overhead
- define the possible resolutions for all templateflow-provided templates

synchon: Just to put it here, `closest_resolution` is used in functions like `load_parcellation` and `load_mask` where one can choose the resolution, but in functions like `get_parcellation` and `get_mask`, we use the target's resolution to resample the data obtained via `load_*` functions.

fraimondo: Junifer needs to able to run without network connections. My take is to scrape on the first time and save in the `junifer_data` directory (if we follow the scrape option).

My issue with hard-coding the resolutions is what might happen if templateflow adds a new resolution. We might need to update the code.

synchon: Templateflow does the caching so should be able to work without network connection. To make this clean and work in a global way, it's better to have something like a `junifer precache` command which downloads required data for datagrabber, parcellations, masks and template spaces.

fraimondo: 100% agree with the `precache`. I though of calling it `prefetch`. And that's why everything that is donwloaded (like parcellations) are stored in the `junifer_data` directory, to make it explicit and permanent across nodes/sessions/users.

In this case, check which of the above solutions works best in your opinion.

synchon: I would want to make a separate PR tackling this specific feature and come back to `get_template` then. What do you think?

fraimondo: Do you mean `precache`? Yes, that's a totally separate PR.

synchon: Yes.

For this PR, the scraping won't be optimal until we have precache.

fraimondo: First we make it work, then we optimise.

synchon: The latest commits should address your concern.

fraimondo (approved):

Issue #243, opened 2023-07-21
[ENH]: Support for template spaces from `templateflow`


synchon:
### Are you requiring a new dataset or marker?

- [X] I understand this is not a marker or dataset request

### Which feature do you want to include?

Adding support for [templateflow](https://www.templateflow.org) will let us access various templates from the `TemplateFlow Archive` which is maintained as a `datalad` superdataset. This will also let users access templates as when they arrive in `templateflow` without much effort.

### How do you imagine this integrated in junifer?

As a new interface that plays well with existing parcellations and masks.

synchon:
On hold until https://github.com/templateflow/templateflow/issues/161 is resolved.

PR #299, opened 2024-02-02
[ENH]: Add support for on-the-fly template space transformation

synchon:
* [x] tests added/passed

This PR adds support for on-the-fly template space transformation in `junifer.data.get_parcellation()` and `junifer.data.get_mask()` enabling merging of multi-space parcellation and mask. The target space is set as per the datagrabber's specification and all of the other assets are transformed to it and then applied.


synchon:
Needs to go after #243

fraimondo (dismissed):

synchon:
@fraimondo Can we merge this?

junifer/data/template_spaces.py
```diff
@@ -121,15 +124,19 @@ def get_template(
...
-        If template in the required resolution is not found.
+        If required template is not found.
 
     """
     # Check for invalid space; early check to raise proper error
     if space not in tflow.templates():
         raise_error(f"Unknown template space: {space}")
 
+    # Check for template type
+    if template_type not in ["T1w", "whole", "gm", "wm", "csf"]:
```
fraimondo, line 137: What do you mean with `"whole"`. Does it come from any toolbox? I know that usually it is named "brain" mask to reference the GM, WM and CSF masks.

fraimondo:
@synchon Just that one question.

junifer/data/template_spaces.py
fraimondo: it should be "brain" then. That's the brain mask.

synchon: Addressed.

PR #302, opened 2024-02-08
[ENH]: Allow `BOLDWarper` to warp between MNI spaces

synchon:
* [x] tests added/passed

This PR allows `BOLDWarper` to warp BOLD data to other MNI spaces like `MNI152Lin`, `MNI152NLin2009cAsym`, `MNI152NLin6Asym` and others.


junifer/preprocess/bold_warper.py
```diff
@@ -49,9 +59,25 @@ class BOLDWarper(BasePreprocessor):
...
-        super().__init__(
-            on="BOLD", required_data_types=["BOLD", self.ref, "Warp"]
-        )
+        # Initialize superclass based on reference
+        if self.ref == "T1w":
+            super().__init__(
+                on="BOLD", required_data_types=["BOLD", self.ref, "Warp"]
+            )
+        elif self.ref in [
+            "MNI152Lin",
```
fraimondo, line 68: Doesn't this depend on the available tranformations? Shall we encapsulate this somewhere?

junifer/preprocess/bold_warper.py
```diff
@@ -110,46 +137,110 @@ def preprocess(
...
-            raise_error(
-                msg=(
-                    "Unknown warp / transformation file extension: "
-                    f"{warp_file_ext}"
-                ),
-                klass=RuntimeError,
-            )
+            # Check pre-requirements for space manipulation
+            if self.ref == input["space"]:
+                logger.info(
```
fraimondo, line 190: This should be a warning if not an error. Why is the BOLDWarper set if you don't need to warp? I prefer an error though.

fraimondo (changes requested):

junifer/preprocess/bold_warper.py
synchon: I've restricted it for now to not be abused, will relax it once it matures.

junifer/preprocess/bold_warper.py
synchon: I don't understand why it should fail with an error, can make it a warning though. But I prefer a logged message instead of a warning in these cases.

junifer/preprocess/tests/test_bold_warper.py
```diff
@@ -106,3 +106,75 @@ def test_BOLDWarper_preprocess(
...
+        The parametrized DataGrabber objects.
+    element : tuple of str
+        The parametrized elements.
+    space : str
+        The parametrized template space to transform to.
+
+    """
+    with datagrabber as dg:
+        # Read data
+        element_data = DefaultDataReader().fit_transform(dg[element])
```
fraimondo, line 172: Can we add a check to see if the warped images are different from the original? It's very basic, but is something.

fraimondo (changes requested):

junifer/preprocess/tests/test_bold_warper.py
synchon: Addressed in the latest commit. Tests pass in juseless.

PR #224, opened 2023-11-07
[BUG]: relative path in `with` block in YAML is interpreted relative from YAML, but copy operation to junifer jobs directory interprets relative to working directory.

synchon:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have the following directory structure:

```
❱ tree                                                                                                                                                   1 !
.
├── 01_queue.sh
├── junifer_extensions
│   ├── juni-farm
│   │   ├── juni_farm
│   │   │   └── datagrabber
│   │   │       ├── hcp_aging.py
│   │   │       ├── hcp_early_psychosis.py
│   │   │       ├── hcp_ya_confounds_cat.py
│   │   │       └── __pycache__
│   │   │           └── hcp_ya_confounds_cat.cpython-310.pyc
│   │   ├── LICENSE
│   │   └── README.md
│   ├── markers
│   └── parcellations
│       ├── data
│       │   ├── README.md
│       │   └── shen_1mm_268_parcellation.nii.gz
│       ├── __pycache__
│       │   └── register_parcellations.cpython-310.pyc
│       └── register_parcellations.py
├── junifer_jobs
│   └── PREPROCESS_HCP1200
└── yamls
    └── hcp_ya.yaml

```

The top level directory here is my working directory. The code in 01_queue.sh is:
```
#!/usr/bin/zsh
# 
# Run this code to create the junifer_jobs directory
junifer queue yamls/${1}.yaml
```

I run `01_queue.sh hcp_ya` which performs `junifer queue` with the `hcp_ya.yaml`

The content of that yaml file is:

```
❱ cat yamls/hcp_ya.yaml                                                                                                                                  1 !
workdir: /tmp

with:
  - ../junifer_extensions/parcellations/register_parcellations.py
  - ../junifer_extensions/juni-farm/juni_farm/datagrabber/hcp_ya_confounds_cat.py


datagrabber:
    kind: MultipleHCP
    tasks:
      - REST1
      - REST2

preprocess:
    kind: fMRIPrepConfoundRemover
    detrend: true
    standardize: true
    strategy:
        wm_csf: full
        global_signal: full
    low_pass: 0.08
    high_pass: 0.01
    masks:
      - compute_epi_mask
      - compute_brain_mask
      - threshold: 0

storage:
  kind: HDF5FeatureStorage
  uri: ../../../data/junifer_storage/hcp_ya/hcp_ya.hdf5


markers:
   
  - name: parc-Finn2015Shen268_marker-empiricalFC
[...]

PR #127, opened 2023-03-30
[BUG]: Storage URI in yaml config does not behave as I would expect.

fraimondo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I use this yaml file:

```
workdir: /tmp

datagrabber:
    kind: DataladAOMICPIOP1
    tasks: "restingstate"
markers:
  - name: Schaefer100x17_FC
    kind: FunctionalConnectivityParcels
    parcellation: Schaefer100x17
    cor_method: correlation
storage: 
  kind: SQLiteFeatureStorage
  uri: ../storage/PIOP1
queue:
  jobname: TestHTCondorQueue
  kind: HTCondor
  env:
    kind: conda
    name: junifer
  mem: 8G
```

I would expect that the storage uri is interpreted relative from the directory where I run `junifer queue`, but it is interpreted relative to the cwd of the process.

### Expected Behavior

I think it will be more intuitive to interpret relative from the directory where i run the `junifer queue`, i.e. my "perceived working directory".

### Steps To Reproduce

Install junifer.

Use this yaml file and run `junifer queue`

```
workdir: /tmp

datagrabber:
    kind: DataladAOMICPIOP1
    tasks: "restingstate"
markers:
  - name: Schaefer100x17_FC
    kind: FunctionalConnectivityParcels
    parcellation: Schaefer100x17
    cor_method: correlation
storage: 
  kind: SQLiteFeatureStorage
  uri: ../storage/PIOP1
queue:
  jobname: TestHTCondorQueue
  kind: HTCondor
  env:
    kind: conda
    name: junifer
  mem: 8G
```

### Environment

```markdown
❱ junifer wtf
junifer:
  version: 0.0.1.dev909
python:
  version: 3.10.6
  implementation: CPython
dependencies:
  click: 8.1.3
  numpy: 1.22.4
  datalad: 0.17.9
  pandas: 1.4.4
  nibabel: 4.0.2
  nilearn: 0.9.2
  sqlalchemy: 1.4.44
  yaml: '6.0'
system:
  platform: Linux-4.19.0-21-amd64-x86_64-with-glibc2.28
environment:
  LC_CTYPE: en_US.UTF-8
[...]

PR #249, opened 2024-02-27
[ENH]: Allow for using `venv` in queue kind

synchon:

Some people do not like to use conda, it would be nice to be able to set the `env` variable to a path of an environment created using [the venv tool](https://docs.python.org/3/library/venv.html) to run a junifer pipeline.

Similar to conda, but with venv.

_No response_

### Anything else to say?

_No response_

### DIFF
--- a/docs/changes/newsfragments/249.feature
+++ b/docs/changes/newsfragments/249.feature
@@ -1,0 +1,1 @@
+Support ``venv`` as environment kind for queueing jobs by `Synchon Mandal`_
--- a/docs/using/queueing.rst
+++ b/docs/using/queueing.rst
@@ -57,7 +57,8 @@ The following parameters are available for HTCondor:
 
   * ``name``: This is the name of the environment to use in case a virtual
     environment is used. It should be the name when ``conda`` is used and
-    the absolute path to the virtualenv when ``venv`` is used.
+    the absolute or relative path to the virtualenv when ``venv`` is used.
+    If relative path is used then it should be relative to the YAML.
 
 * ``mem``: Memory to be used by the job. It must be provided as a string with
   the units (e.g. ``2GB``).
--- a/docs/using/queueing.rst
+++ b/docs/using/queueing.rst
@@ -52,11 +52,12 @@ The following parameters are available for HTCondor:
   * ``kind``: This is the kind of virtual environment to use:
 
     * ``conda``
-    * ``virtualenv`` (not yet supported)
+    * ``venv``
     * ``local`` (no virtual environment)
 
   * ``name``: This is the name of the environment to use in case a virtual
-    environment is used.
+    environment is used. It should be the name when ``conda`` is used and
+    the absolute path to the virtualenv when ``venv`` is used.
 
 * ``mem``: Memory to be used by the job. It must be provided as a string with
   the units (e.g. ``2GB``).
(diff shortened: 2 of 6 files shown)


fraimondo (approved):

fraimondo:
You can ask @LeSasse how does he feel about this one too...

docs/using/queueing.rst
```diff
@@ -52,11 +52,12 @@ The following parameters are available for HTCondor:
...
 
     * ``conda``
-    * ``virtualenv`` (not yet supported)
+    * ``venv``
     * ``local`` (no virtual environment)
 
   * ``name``: This is the name of the environment to use in case a virtual
-    environment is used.
+    environment is used. It should be the name when ``conda`` is used and
+    the absolute path to the virtualenv when ``venv`` is used.
```
LeSasse, line 60: no support for a relative path? What is the main rationale behind this?

docs/changes/newsfragments/249.feature
```diff
@@ -0,0 +1 @@
+Support ``venv`` as environment kind for queueing jobs by `Synchon Mandal`_
```
LeSasse, line 1: nice!

junifer/api/res/run_venv.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
LeSasse, line 1: hard coded shell? I suppose this will change in a separate PR?

junifer/api/res/run_venv.sh
```diff
@@ -0,0 +1,22 @@
...
+
+if [ $# -lt 2 ]; then
+    echo "This script is meant to run a command within a python environment"
+    echo "It needs at least 2 parameters."
+    echo "The first one must be the environment name."
```
LeSasse, line 6: the environment path in this case i suppose, might be worth mentioning here specifically

LeSasse (commented):
Overall, this PR is the real MVP. Just a few questions for clarification from my side.

docs/using/queueing.rst
fraimondo: The main question is: relative to what? The YAML? the CWD? We had this issue in #224, #127, #191.

If we can answer this, then we can also use a relative path.

junifer/api/res/run_venv.sh
fraimondo: I think this is the case. We need to provide `run_venv.zsh`, `run_conda.zsh`, etc.

docs/using/queueing.rst
LeSasse: i agree. Absolute will keep it simple.

junifer/api/res/run_venv.sh
fraimondo: +1

docs/using/queueing.rst
fraimondo: I think relative paths are simpler, problem is always the same: determining the "relative to" to compute an absolute path and fix it in the yaml.

junifer/api/res/run_venv.sh
synchon: Yes this will change in #273 which will require a new option in the YAML.

docs/using/queueing.rst
synchon: We have been following the "relative to YAML" idea and I updated it to reflect that.

junifer/api/res/run_venv.sh
synchon: Updated it to make it clear.

LeSasse (approved):
LGTM, merge once tests pass I suppose.

PR #111, opened 2022-10-29
WIP: Preprocess

fraimondo:
* [ ] description of feature/fix
* [x] tests added/passed
* [ ] add an entry to the [latest changes](../docs/changes/latest.inc)

PR #310, opened 2024-03-07
[ENH]: Improve `BasePreprocessor`

synchon:
* [x] tests added/passed

This PR refactors `preprocess` method of preprocessors to not return "data type" and only return the preprocessed input data. Having to return "data type" forced a concrete preprocessor to only operate on a single "data type" (like `BOLDWarper` introduced in #267) and not allow the `on` parameter to be exposed to the user (as requested in #301). If a concrete preprocessor adds new "data type" like `BOLD_mask` in the "junifer data object" as `fMRIPrepConfoundRemover` (introduced in #111) does, it will be handled as usual with no changes.

A preprocessor should not create new "data types" (which was allowed earlier and hence the restriction) but only create and add "helper data types" like `BOLD_mask` which happens via `extra_input` of the `preprocess` method.

### DIFF
--- a/junifer/preprocess/confounds/fmriprep_confound_remover.py
+++ b/junifer/preprocess/confounds/fmriprep_confound_remover.py
@@ -600,7 +600,7 @@ class fMRIPrepConfoundRemover(BasePreprocessor):
         self,
         input: Dict[str, Any],
         extra_input: Optional[Dict[str, Any]] = None,
-    ) -> Tuple[str, Dict[str, Any]]:
+    ) -> Dict[str, Any]:
         """Preprocess.
 
         Parameters
@@ -613,13 +613,10 @@ class fMRIPrepConfoundRemover(BasePreprocessor):
 
         Returns
         -------
-        str
-            The key to store the output in the Junifer Data object.
         dict
-            The computed result as dictionary. This will be stored in the
-            Junifer Data object under the key ``data`` of the data type.
+            The computed result as dictionary.
 
         """
         self._validate_data(input, extra_input)
         input["data"] = self._remove_confounds(input, extra_input=extra_input)
-        return "BOLD", input
+        return input
--- a/junifer/preprocess/confounds/fmriprep_confound_remover.py
+++ b/junifer/preprocess/confounds/fmriprep_confound_remover.py
@@ -6,7 +6,6 @@
 # License: AGPL
 
 from typing import (
-    TYPE_CHECKING,
     Any,
     ClassVar,
     Dict,
@@ -19,17 +18,13 @@ from typing import (
 
 import numpy as np
 import pandas as pd
+from nilearn import image as nimg
 from nilearn._utils.niimg_conversions import check_niimg_4d
-from nilearn.image import clean_img
 
 from ...api.decorators import register_preprocessor
 from ...data import get_mask
 from ...utils import logger, raise_error
 from ..base import BasePreprocessor
-
-
-if TYPE_CHECKING:
-    from nibabel import MGHImage, Nifti1Image, Nifti2Image
 
 
 FMRIPREP_BASICS = {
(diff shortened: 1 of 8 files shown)


fraimondo:
I think I have an issue with this PR.

1) How is the `BOLD_mask` kept even before the PR? I check the `fMRIPrepConfoundRemover` and all we do is add it to the `extra_input`. But how is this kept???? Is this actually working???

2) Not allowing the preprocessor to "create new datatypes" is a bit strong. Indeed there's no difference between "new data types" and "helper data types".

synchon:
> 1. How is the `BOLD_mask` kept even before the PR? I check the `fMRIPrepConfoundRemover` and all we do is add it to the `extra_input`. But how is this kept???? Is this actually working???

Since the `input` is not copied for `extra_input`, it's kept but if we copy it like we do for markers, it won't be kept.

> 2. Not allowing the preprocessor to "create new datatypes" is a bit strong. Indeed there's no difference between "new data types" and "helper data types".

The reason is because `BOLD` or `T1w` would be a "data type" for me but not `BOLD_mask`. But, we are on the same page with our understanding.

synchon:
> > 1. How is the `BOLD_mask` kept even before the PR? I check the `fMRIPrepConfoundRemover` and all we do is add it to the `extra_input`. But how is this kept???? Is this actually working???
> 
> Since the `input` is not copied for `extra_input`, it's kept but if we copy it like we do for markers, it won't be kept.

We could have the `extra_input` be returned from `preprocess` method to make it explicit though. What do you think?

fraimondo:
Check this out:

This is how we "add" the `BOLD_mask`:

https://github.com/juaml/junifer/blob/a7bd0f6449fe9c2298501ce2d6a27349ab850190/junifer/preprocess/confounds/fmriprep_confound_remover.py#L575-L580

But this is how we treat the input in the `_fit_transform` function:

https://github.com/juaml/junifer/blob/a7bd0f6449fe9c2298501ce2d6a27349ab850190/junifer/preprocess/base.py#L173-L192


So basically we have `out`, which is `input`, which is also `extra_input`. We then `pop` the `type`, modify the `extra_input` and re-add the `type` to `out`.

This is madness. My madness, but madness still.

I don't know how to tackle this to make it proper.

Problems we currently have:
1) We are splitting the `input`, which is also the `out` variable in two, for no reason.
2) We are giving the `preprocess` function all the data (`input`), even if they do not need it.
3) We are "adding" new "data types" by modifying the `extra_input` variable.

A more declarative way would be to:

1) Only pass the `input` and the relevant `extra_input` to the preprocess function (not all the data)
2) Use the `returned` values to update the `out`. Maybe by getting a dictionary of key/value pairs?

fraimondo:
Perfect!

synchon:
> > A more declarative way would be to:
> > 
> > 1. Only pass the input and the relevant extra_input to the preprocess function (not all the data)
> > 2. Use the returned values to update the out. Maybe by getting a dictionary of key/value pairs?
> 
> I agree with the steps and it's similar to what I proposed in the earlier comment.
> 
> * In the beginning of `_fit_transform`, we could copy the `input` to `out` like we do for DataReader so that we don't lose anything.
> * Then from `preprocess` we return the first value as the `input` dict (as we do now).
> * The second value from `preprocess` could be "helper data type(s)" dict or None and we check for it in `_fit_transform`. If we have it as None, we don't do anything, else we update `out`.

@fraimondo The latest commits should implement this.

junifer/preprocess/confounds/fmriprep_confound_remover.py
```diff
@@ -528,100 +513,78 @@ def _validate_data(
...
+                    }
+                }
+        # Clean image
+        logger.info("Cleaning image using nilearn")
         logger.debug(f"\tdetrend: {self.detrend}")
         logger.debug(f"\tstandardize: {self.standardize}")
         logger.debug(f"\tlow_pass: {self.low_pass}")
         logger.debug(f"\thigh_pass: {self.high_pass}")
-        clean_bold = clean_img(
+        input["data"] = nimg.clean_img(
```
fraimondo, line 579: I still do prefer to have it coded without explicit side-effects as it was before.

fraimondo (approved):
Tests are not passing yet, but code is OK from my side.

junifer/preprocess/confounds/fmriprep_confound_remover.py
synchon: I debated on this in my head for some time and decided to go with this due to the reduced complexity of having everything in `preprocess` and the need to now return `BOLD_mask` via the `extra_input` which is much more explicit in this.

synchon: If you feel strongly about it and don't see an upside with the change, I can revert it with the required change.

synchon:
> Tests are not passing yet, but code is OK from my side.

Was a network issue for downloading assets, nothing from our side. Let's wait and see.

PR #311, opened 2024-03-14
Improve external dependency handling of `PipelineStepMixin`

synchon:
* [x] tests added/passed

This PR improves the external dependency handling of `PipelineStepMixin` derived objects like Markers and Preprocessors. It adds a new class variable called `_CONDITIONAL_DEPENDENCIES` which interact with `_DEPENDENCIES` and `_EXT_DEPENDENCIES`. This enables easy validation and usage of objects which have multiple implementations like `ReHo`, `ALFF` and `BOLDWarper`. It also sets a nice foundation for similar objects in the future as will be implemented in #161 and #301. It also refactors internal code wherever required.


fraimondo (changes requested):
Can you also update `docs/extending/markers.rst`? Maybe create a small section on "dependencies" and then refer to that?

This way, we will document everything that is happening on this PR (which was also a bit undocumented before)

synchon:
> Can you also update `docs/extending/markers.rst`? Maybe create a small section on "dependencies" and then refer to that?
> 
> This way, we will document everything that is happening on this PR (which was also a bit undocumented before)

I'm already updating the docs for the release at https://github.com/juaml/junifer/tree/update/improve-docs . Will address your suggestions there along with a few more things that I noticed aren't there.

fraimondo (approved):

PR #301, opened 2024-03-15
[ENH]: Preprocessing step to warp any image to any standard space

synchon:

So far, the `BOLDWarper` allows to warp the BOLD image.

However, given that we have multiple MNI spaces, it would be a good idea to have a `SpaceWarper` preprocessor (or similar) that exposes the `on` parameter to the user.

A `SpaceWarper` or similar class.

_No response_

### Anything else to say?

_No response_


junifer/preprocess/warping/space_warper.py
```diff
@@ -0,0 +1,258 @@
...
+                    f"data_warped_from_{input['space']}_to_"
+                    f"{self.reference}.nii.gz"
+                )
+
+                logger.debug(
+                    f"Using ANTs to warp data "
+                    f"from {input['space']} to {self.reference}"
+                )
+                # Set antsApplyTransforms command
+                apply_transforms_cmd = [
```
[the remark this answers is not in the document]
synchon: The `ANTsWarper` works and behaves in a different way as it also uses `ResampleImage`, I wouldn't want to keep this there. We can create a new helper class with an entry in `_EXT_DEPENDENCIES`.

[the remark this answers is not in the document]
fraimondo: `ANTSTemplateWarper`?

junifer/preprocess/warping/space_warper.py
```diff
@@ -0,0 +1,258 @@
...
+        """Get valid data types for input.
+
+        Returns
+        -------
+        list of str
+            The list of data types that can be used as input for this
+            preprocessor.
+
+        """
+        return ["T1w", "BOLD"]
```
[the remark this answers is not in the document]
fraimondo: For the moment, all of this but `BOLD_confounds`: https://juaml.github.io/junifer/main/understanding/data.html#data-types

junifer/preprocess/warping/space_warper.py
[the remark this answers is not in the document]
synchon: Sounds good to me.

junifer/preprocess/warping/space_warper.py
```diff
@@ -0,0 +1,198 @@
...
+@register_preprocessor
+class SpaceWarper(BasePreprocessor):
+    """Class for warping data to other template spaces.
+
+    Parameters
+    ----------
+    using : {"fsl_native", "ants_native", "ants_template}
+        Implementation to use for warping:
+
+        * "fsl_native" : Use FSL's ``applywarp`` for native space warping
```
fraimondo, line 30: Doesn't this need to happen automatically? I mean, why do the user need to change one parameter more than just the `reference`?

If `reference` is `"T1w"`, then we use the native one, if not, the template one.

I think this is one of the places where the user should not need to change more than the `reference` parameter.

junifer/preprocess/warping/space_warper.py
```diff
@@ -0,0 +1,198 @@
...
+                raise_error(
+                    (
+                        "Skipped warping as the data is in "
+                        f"{self.reference} space which would mean that you "
+                        "can remove the SpaceWarper from the preprocess step."
+                    ),
+                    klass=RuntimeError,
+                )
+            else:
+                input = ANTsTemplateWarper().preprocess(
```
fraimondo, line 193: This also has a conflict with the `using` parameter. It does not consider the `using` here, so `fsl` or `ants` is the same here.

fraimondo (changes requested):

junifer/preprocess/warping/space_warper.py
synchon: From our conversation above:
>> Gathering from your replies but quoting only this, I definitely support a single parameter; let's finalise it as method or using which goes in the constructor of every marker / preprocessor having this behaviour. I would want this parameter to be positional and not optional with no default value and no value as "auto". This would keep the metadata consistent. Now, after that the check happens as you already described above.
>
>100% agree.

I wouldn't want to make exception for any component. I agreed with you for `BOLDWarper` auto-detecting considering we had small scope. For this I disagree and my rationale is quoted.

> I think this is one of the places where the user should not need to change more than the reference parameter.

IMO it wouldn't hurt the user to add one extra line when in return it would give us complete knowledge of the component. I don't want to trade clarity for saving one line in the YAML.

junifer/preprocess/warping/space_warper.py
synchon: Unfortunately, there is no other current approach for warping templates in junifer so having a forced check for `using` doesn't really make sense. Of course we can make it but it doesn't give anything extra in return. I can make the base `if...else` stronger if that eases this out.

synchon:
@fraimondo Shall we merge this?

junifer/preprocess/warping/space_warper.py
```diff
@@ -0,0 +1,192 @@
...
+            If ``extra_input`` is None when transforming to native space
+            i.e., using ``"T1w"`` as reference.
+        RuntimeError
+            If the data is in the correct space and does not require
+            warping.
+
+        """
+        logger.info(f"Warping to {self.reference} space using SpaceWarper")
+        # Transform to native space
+        if self.using in ["fsl", "ants"] and self.reference == "T1w":
```
fraimondo, line 154: I'm missing one condition here (which should raise an error)

`using == "fsl"` and `reference != T1w"`

junifer/preprocess/warping/space_warper.py
```diff
@@ -0,0 +1,192 @@
...
+        elif self.using == "ants" and self.reference != "T1w":
+            # Check pre-requirements for space manipulation
+            if self.reference == input["space"]:
+                raise_error(
+                    (
+                        "Skipped warping as the data is in "
+                        f"{self.reference} space which would mean that you "
+                        "can remove the SpaceWarper from the preprocess step."
+                    ),
+                    klass=RuntimeError,
```
fraimondo, line 183: If this is an error, then the message should not be "skipped...", "can remove...".

Other option is to make it a warning.

Following the strict pipeline policy, we should keep it an error and change the message indicating that the user MUST remove that preprocessing step.

fraimondo (changes requested):
Can we also add some tests for the errors?

fraimondo (approved):

PR #273, opened 2024-03-20
[ENH]:  Specify which shell to use when using `junifer queue`

synchon:

I would like to be able to specify the shell that junifer uses to submit jobs. For some reason when using bash I could not get jobs to work. When using zsh by manually changing it, it seemed to work (maybe something from with my bash environment, maybe magic or maybe coincidence). Either way I think this could be a desirable feature. What do you think? 

as a param that prints different text

_No response_

### Anything else to say?

_No response_


synchon:
@LeSasse Can you check if the branch solves your issue?

junifer/api/queue_context/htcondor_adapter.py
```diff
@@ -178,7 +188,7 @@ def _check_collect(self, collect: str) -> str:
     def pre_run(self) -> str:
         """Return pre-run commands."""
         fixed = (
-            "#!/bin/bash\n\n"
+            f"#!/bin/{self._shell}\n\n"
```
LeSasse, line 191: would it be better to do `/bin/env {shell}`?

junifer/api/res/run_conda.zsh
```diff
@@ -0,0 +1,23 @@
+#!/bin/zsh
```
LeSasse, line 1: would it be viable to have the `res` as a plain `txt` containing a text with variable placeholders that get replaced by fill values for a specific shell or are the scripts for the different shell vastly different? just wondering if its desirable avoiding 1 script per shell.

junifer/api/res/run_venv.zsh
```diff
@@ -0,0 +1,22 @@
+#!/bin/zsh
```
LeSasse, line 1: nice

junifer/api/queue_context/htcondor_adapter.py
synchon: The standard advice that I see in the wild is to use `#!/bin/<shell>`. Is there a particular reason for your suggestion?

junifer/api/res/run_conda.zsh
synchon: We would need the file to be created on-demand for that and not have it as a file. I wanted a single file but for simplicity made it separate.

junifer/api/queue_context/htcondor_adapter.py
LeSasse: Running a command through /usr/bin/env has the benefit of looking for whatever the default version of the program is in your current environment.

This way, you don't have to look for it in a specific place on the system, as those paths may be in different locations on different systems. As long as it's in your path, it will find it.

But at the end of the day it isnt really going to make a difference.

synchon: That's a fair argument if you want to have control over the shell used. Would you say we extend it to `run_{conda,venv}.{bash,zsh}` as well?

LeSasse: Yes, i think so, but didnt wanna duplicate the comment.

LeSasse:
> @LeSasse Can you check if the branch solves your issue?

Will check it

fraimondo (approved):

LeSasse:
> @LeSasse Can you check if the branch solves your issue?

I think this branch needs a rebase: the dag files still use single quotes and submitting therefore fails.

synchon:
> > @LeSasse Can you check if the branch solves your issue?
> 
> I think this branch needs a rebase: the dag files still use single quotes and submitting therefore fails.

I see that the quotes are proper in `dag`, can you try rebasing your local branch?

LeSasse:
> > > @LeSasse Can you check if the branch solves your issue?
> > 
> > 
> > I think this branch needs a rebase: the dag files still use single quotes and submitting therefore fails.
> 
> I see that the quotes are proper in `dag`, can you try rebasing your local branch?

thats quite odd because i actually cloned and checked out this branch just minutes before that comment, and after i rebased it locally it worked, but before, no.

LeSasse:
Tested this now and it works.

PR #308, opened 2024-04-03
[ENH]: Improving DataGrabber `patterns`

synchon:
### Discussed in https://github.com/juaml/junifer/discussions/307

<div type='discussions-op-text'>

<sup>Originally posted by **synchon** February 26, 2024</sup>
With the introduction of template space definitions in #252 and #268, the use of `PatternDataladDataGrabber` via YAML is broken as of `9316bca` due to space awareness. The `space` key is added in `get_item` method of the DataGrabber and not defined in the constructor. The `space` key is important for every other component downstream and not having that breaks the pipeline. The solution I propose is not backwards compatible as it re-defines the structure of the `patterns` parameter, but in my opinion the cleanest and leanest way:

### Python:
```python
patterns = {
    "BOLD": "...",
    "T1w": "...",
}
```

### YAML:
```yaml
patterns:
  BOLD: "..."
  T1w: "..."
```

### Python:
```python
patterns = {
    "BOLD": {
        "pattern": "...",
        "space": "...",
    },
    "T1w": {
        "pattern": "...",
        "space": "...",
    },
}
```

### YAML:
```yaml
patterns:
  BOLD:
    pattern: "..."
    space: "..."
  T1w:
    pattern: "..."
    space: "..."
```

I'm happy to discuss alternatives and adapt my solution.</div>

### DIFF
--- a/docs/changes/newsfragments/308.enh
+++ b/docs/changes/newsfragments/308.enh
@@ -1,0 +1,1 @@
+Improve :class:`.PatternDataGrabber` and :class:`.PatternDataladDataGrabber`'s ``patterns`` to enable ``space``, ``format``, ``mask_item`` and other metadata description handling via YAML by `Synchon Mandal`_


fraimondo (approved):

LeSasse (approved):
LGTM

PR #321, opened 2024-04-04
[FIX] change files to executables.

fraimondo:
* [x] fix issue with executable flags in some files
* [x] tests added/passed


docs/changes/newsfragments/321.fix
synchon, line 1: Maybe generalise it coz it affects more than ANTs' scripts?

synchon (approved):

PR #323, opened 2024-04-16
[ENH]: Add `junifer list-elements` for listing out elements of an YAML's  DataGrabber

synchon:

I think it would be fantastic if junifer had a `list-elements` command line function which can be run on a yaml file with a specific datagrabber, and that will then print out all available elements in the format that junifer likes for subsetting the element keys using the `--element elements.txt` command line option for queue.

The workflow i imagine would have the advantage that i can very easily discover elements and subset them for testing using arbitrary command line filters, and then write them into a file for queing i.e.:

`junifer list-elements <yaml> | tail -2 > elements.txt`

as above

_No response_

### Anything else to say?

_No response_

### DIFF
--- a/docs/changes/newsfragments/323.feature
+++ b/docs/changes/newsfragments/323.feature
@@ -1,0 +1,1 @@
+Add ``junifer list-elements`` to list out available elements for a DataGrabber based on filtering via ``--element`` by `Synchon Mandal`_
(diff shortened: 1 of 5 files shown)


fraimondo:
Does this work?
```
junifer list-elements <yaml> | tail -2 > elements.txt
```

LeSasse (approved):
LGTM

PR #344, opened 2024-05-16
[ENH]: Add support for `BrainPrint`

synchon:
* [x] tests added/passed

This PR adds support for `BrainPrint` marker.

### DIFF
--- a/junifer/markers/tests/test_brainprint.py
+++ b/junifer/markers/tests/test_brainprint.py
@@ -1,0 +1,47 @@
+"""Provide tests for BrainPrint."""
+
+# Authors: Synchon Mandal <s.mandal@fz-juelich.de>
+# License: AGPL
+
+import socket
+
+import pytest
+
+from junifer.datagrabber import DataladAOMICID1000
+from junifer.datareader import DefaultDataReader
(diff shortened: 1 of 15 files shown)


junifer/markers/tests/test_brainprint.py
```diff
@@ -0,0 +1,47 @@
...
+    """Test BrainPrint validate()."""
+    marker = BrainPrint()
+    assert set(marker.validate(["FreeSurfer"])) == {"scalar_table", "vector"}
+
+
+@pytest.mark.skipif(
+    _check_freesurfer() is False, reason="requires FreeSurfer to be in PATH"
+)
+@pytest.mark.skipif(
+    socket.gethostname() != "juseless",
```
fraimondo, line 32: Why only for juseless?

fraimondo (approved):

junifer/markers/tests/test_brainprint.py
synchon: The data takes quite some time to fetch and the computation is better half of an hour, so would exceed the limitations of the gh-actions standard runners. We can pull it off on gh-actions if we have a small testing dataset providing FreeSurfer derivatives.

PR #345, opened 2024-05-21
[ENH]: Unix path expansion directives support for `PatternDataGrabber`

synchon:
* [x] tests added/passed

This PR adds support for specifying Unix path expansion directives like `[], [!]` for `PatternDataGrabber` patterns. This is an improvement which is needed for supporting `FreeSurfer` data type. For example, `fsaverage` directory is at the same level as subjects like `sub-0001` for AOMIC datasets and `[!f]` allows to not glob the `fsaverage` directory without overriding `get_elements()` method. As the `.mgz` and other outputs of `FreeSurfer` are not unique by subject (and/or task / session) prefix, globbing doesn't work as usual (at least for the AOMIC datasets). This also enables one to have the pattern: `.../T1.mg[z]` to make it work.

### DIFF
--- a/docs/changes/newsfragments/345.feature
+++ b/docs/changes/newsfragments/345.feature
@@ -1,0 +1,1 @@
+Allow Unix path expansion directives to be used in :class:`.PatternDataGrabber` ``patterns`` by `Synchon Mandal`_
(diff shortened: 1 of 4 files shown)


synchon:
Needs to go before #344

fraimondo (changes requested):
Can you also improve the docs to reflect this change?

fraimondo (approved):

PR #136, opened 2024-06-12
[ENH]: Add test for `JuniferNiftiSpheresMasker` against nilearn's version

synchon:

We reimplemented (by copying) a `JuniferNiftiSpheresMasker` that has the same functionality of `NiftiSpheresMasker` but allows to change the aggregation function.

The tests are also a copy of Nilearn's one.

If `agg_function = 'mean'`, then they should both behave equally. I would like to have a test so we can corroborate this. They should be numerically equal.

One more test in `junifer.external.nilearn.tests`

_No response_

### Anything else to say?

_No response_


synchon:
#347 needs to go in before this as that sets the proper nilearn version we need for this, #333 and #335

fraimondo (approved):

PR #348, opened 2024-06-14
[ENH]: Introduce `JuniferConnectivityMeasure`

synchon:
* [x] tests added/passed

This PR introduces `JuniferConnectivityMeasure`, tailored `nilearn.connectome.ConnectivityMeasure` for junifer. This sets `sklearn.covariance.EmpiricalCovariance` as the default covariance estimator and `correlation` as the default connectivity matrix kind, which is what junifer users mostly do. It also sets the ground to add custom connectivity matrix kinds and correlation metrics. Required by #333 and #335 as of now.


fraimondo (approved):
This will require a new major I think.

PR #361, opened 2024-08-29
[ENH]: Introduce new `cli` sub-package

synchon:
* [x] tests added/passed

This PR refactors the CLI components and splits the `api` sub-package into a new `cli` sub-package for improved intent and maintenance. It also reduces the side-effects introduced by CLI components, leaking into the public API components and plays well with lazy imports.


fraimondo:
I like what this PR is doing, but I think we are still failing to do it properly (same mistake as before).

The `cli` module should take care of click and all the command-line-interface parsing stuff.
The `api` module is the application programming interface module. 

The first module should parse/interpret the cli commands and call the second module, which is the one available for the user in case they decide to write python code to use junifer.

synchon:
> I like what this PR is doing, but I think we are still failing to do it properly (same mistake as before).
> 
> The `cli` module should take care of click and all the command-line-interface parsing stuff. The `api` module is the application programming interface module.
> 
> The first module should parse/interpret the cli commands and call the second module, which is the one available for the user in case they decide to write python code to use junifer.

I see what you mean, makes sense to me. I didn't wanna change a lot of stuff before an initial review so will make the further changes now.

fraimondo:
> > I like what this PR is doing, but I think we are still failing to do it properly (same mistake as before).
> > The `cli` module should take care of click and all the command-line-interface parsing stuff. The `api` module is the application programming interface module.
> > The first module should parse/interpret the cli commands and call the second module, which is the one available for the user in case they decide to write python code to use junifer.
> 
> I see what you mean, makes sense to me. I didn't wanna change a lot of stuff before an initial review so will make the further changes now.

It was my original intention, but things got too mixed.

synchon:
> > > I like what this PR is doing, but I think we are still failing to do it properly (same mistake as before).
> > > The `cli` module should take care of click and all the command-line-interface parsing stuff. The `api` module is the application programming interface module.
> > > The first module should parse/interpret the cli commands and call the second module, which is the one available for the user in case they decide to write python code to use junifer.
> > 
> > 
> > I see what you mean, makes sense to me. I didn't wanna change a lot of stuff before an initial review so will make the further changes now.
> 
> It was my original intention, but things got too mixed.

There, should be cleaner now. Good for review.

fraimondo (approved):

PR #367, opened 2024-09-26
[ENH]: Allow for external modules to import other files

fraimondo:

I am currently working on an extension in juni-farm which requires another .py file to be imported. When I add it to the `with` section of the yaml, even though both files are in there, it fails with a `ModuleNotFoundError`.

We originally though of having extensions be a single file, but now things have grown and it's not suitable anymore.

For example, in `my_extension_b.py`, we can have something like 

```python
from my_extension_a import CoolDataGrabber

@registerdatagrabber
class SuperCoolDataGrabber(CoolDataGrabber):
...
```

The solution is quite simple, for this to work, we need to add the path of each module to `sys.path`, so all imports work properly.

This creates an issue with the YAML file in the case of the `queue` command being inconsistent with the `run` one. 

On the `run` scenario, the module is ran from it's original location, where both files are present. In the `queue` scenario, only the elements in the `with` statement are copied to the `junifer_jobs` respective directory. 

In order to work, it needs to modifications in junifer:

1) Before importing any module in the with statement, add it's parent directory to `sys.path`
2) After loading the module, check if the module has a `junifer_module_deps` function. If so, call it and add all of this files to the `with` section of the parsed yaml. This will create the full yaml required for the `queue` command to work.

_No response_

### Anything else to say?

_No response_

### DIFF
--- a/docs/extending/extension.rst
+++ b/docs/extending/extension.rst
@@ -37,6 +37,7 @@ This is the ideal place to include ``junifer`` extensions.
    ``junifer`` about the dependencies of the module:
 
    .. code-block:: python
+
       def junifer_module_deps() -> List[str]:
         """Return the dependencies of the module.
 
--- a/docs/extending/extension.rst
+++ b/docs/extending/extension.rst
@@ -28,8 +28,24 @@ This is the ideal place to include ``junifer`` extensions.
 
 .. important::
 
-   Some ``junifer`` commands will not consider files imported from files included
-   in the ``with`` statement. If ``my_file.py`` imports ``my_other_file.py``,
-   some of the ``junifer`` commands will not consider ``my_other_file.py``. Either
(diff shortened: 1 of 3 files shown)


synchon (approved):
Should be good to merge once the docs are updated and the CI checks pass.

docs/extending/extension.rst
```diff
@@ -28,8 +28,24 @@ This is the ideal place to include ``junifer`` extensions.
...
+   so, we need to include a special function in ``my_file.py`` which tells
+   ``junifer`` about the dependencies of the module:
+
+   .. code-block:: python
+      def junifer_module_deps() -> List[str]:
+        """Return the dependencies of the module.
+
+        Returns
+        -------
+        List[str]
```
synchon, line 45: Slight nitpick: `list of str`

fraimondo: done

PR #374, opened 2024-10-11
[CI]: Adopt custom CI image for junifer docs

synchon:
* [x] tests added/passed

This PR adopts custom CI image for building junifer docs and docs preview.

### DIFF
--- a/.github/workflows/docs-preview.yml
+++ b/.github/workflows/docs-preview.yml
@@ -18,8 +18,12 @@ concurrency: preview-${{ github.ref }}
 jobs:
   build-docs:
     runs-on: ubuntu-latest
+    # Use custom docs CI image
     container: ghcr.io/juaml/junifer-docs:main
     steps:
+    - name: Check ANTs
+      run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/docs-preview.yml
```diff
@@ -18,7 +18,24 @@ concurrency: preview-${{ github.ref }}
...
+    # Use custom docs CI image
+    container: ghcr.io/juaml/junifer-docs:main
     steps:
+    - name: Check ANTs
+      run: |
+        echo "Using ANTs : $(antsRegistration --version)"
+    - name: Install system dependencies
+      run: |
+        # actions/checkout fails to identify Git as it's not in the CI image
+        apt-get -qq update && \
```
fraimondo, line 30: Why can't we have this already in the image? Same as the "home dir" steps. Ideally, this image should be also usable externally.

synchon: Currently I couldn't find a way to put it inside the image as the build kept failing. I wanted the same as well so it's usable, but I guess it'll take some trial and error to fix this mess that gh-action has.

PR #387, opened 2024-10-29
[ENH]: Introduce data registries

synchon:
* [x] tests added/passed

This PR introduces new singleton classes: `ParcellationRegistry`, `CoordinatesRegistry` and `MaskRegistry` for centralizing data registries used in the codebase, easing encapsulation. It also replaces helper data functions: `{get,load,list,register}_{parcellation,coordinates,mask}` with new "dynamically dispatched" functions like: `{get,load,list,register,deregister}_data` which act as helpers for the data registries.


fraimondo (approved):

fraimondo:
There's an issue with the 3.12 tests, regarding FSL. Once this is fixed and CI is green, can be merged.

synchon:
> There's an issue with the 3.12 tests, regarding FSL. Once this is fixed and CI is green, can be merged.

It's something specific to FSL `susan` and not any other FSL tool we use, tried fixing in the CI image but doesn't seem to work as it somehow needs to be hardcoded in `.profile` I guess. As the problem is unrelated to this PR, I can tackle it in another one; since I need this PR for the following ones.

synchon:
> > There's an issue with the 3.12 tests, regarding FSL. Once this is fixed and CI is green, can be merged.
> 
> It's something specific to FSL `susan` and not any other FSL tool we use, tried fixing in the CI image but doesn't seem to work as it somehow needs to be hardcoded in `.profile` I guess. As the problem is unrelated to this PR, I can tackle it in another one; since I need this PR for the upcoming ones.

@fraimondo Got the solution in #383 but it changes the CI testing env so we can merge this and then get that in and continue, what do you think?

fraimondo:
> @fraimondo Got the solution in #383 but it changes the CI testing env so we can merge this and then get that in and continue, what do you think?

As long as the end product is green, I'm ok with either way.

PR #391, opened 2024-11-07
[ENH]: Refactor `Singleton` as a metaclass

fraimondo:
Refactor the Singleton so it stays a metaclass, allowing to reset the instances. 

This is needed if we want to test some elements independently and "reset" them between unit tests. E.g. the `WorkDirManager`


fraimondo:
@synchon need this in order to merge #389

synchon (approved):

PR #390, opened 2024-11-08
[ENH]: Update `Warp` datatype schema and logic

synchon:
1) Modify the `Warp` datatype to be a "list" of mappings from / to spaces.
2) Add a `"warper"` key to the `Warp` datatype indicating which warper was used: `fsl` or `ants` for the moment.
3) Modify the `SpaceWarper` to allow to `"auto"` select the warper based on the dataset. This should require both FSL and ANTs as external dependencies. This only works if reference is `"T1w"`: `"auto"` is not defined for junifer internal transforms (i.e. between template spaces)
4) Modify coordinates/masks/parcellations so they check for the right transform and use the "warper" parameter accordingly.

### DIFF
--- a/docs/extending/dependencies.rst
+++ b/docs/extending/dependencies.rst
@@ -39,7 +39,7 @@ and others who use it will thank you.
 Handling external dependencies from toolboxes
 ---------------------------------------------
 
-You can also specify dependencies of external toolboxes like AFIN, FSL and ANTs,
+You can also specify dependencies of external toolboxes like AFNI, FSL and ANTs,
 by having a class attribute like so:
 
 .. code-block:: python
@@ -87,6 +87,10 @@ that it shows the problem a bit better and how we solve it:
                 "using": "ants",
                 "depends_on": ANTsWarper,
             },
+            {
+                "using": "auto",
+                "depends_on": [FSLWarper, ANTsWarper],
+            },
         ]
 
         def __init__(
@@ -100,12 +104,19 @@ Here, you see a new class attribute ``_CONDITIONAL_DEPENDENCIES`` which is a
 list of dictionaries with two keys:
 
 * ``using`` (str) : lowercased name of the toolbox
-* ``depends_on`` (object) : a class which implements the particular tool's use
+* ``depends_on`` (object or list of objects) : a class or list of classes which \
+  implements the particular tool's use
 
 It is mandatory to have the ``using`` positional argument in the constructor in
 this case as the validation starts with this and moves further. It is also
 mandatory to only allow the value of ``using`` argument to be one of them
 specified in the ``using`` key of ``_CONDITIONAL_DEPENDENCIES`` entries.
+
+For some cases, like we have here, it might be worth to have ``"auto"`` for
+``"using"`` which eases the choice of a particular tool by the user and
+instead lets ``junifer`` do it automatically. In that case, ``"depends_on"``
+needs to be a list of the tool implementation classes. This also requires the
+user to have all the tools in the ``PATH``.
 
 For brevity, we only show the ``FSLWarper`` here but ``ANTsWarper`` looks very
 similar. ``FSLWarper`` looks like this (only the relevant part is shown here):
--- a/junifer/data/coordinates/_coordinates.py
+++ b/junifer/data/coordinates/_coordinates.py
@@ -311,7 +311,7 @@ class CoordinatesRegistry(BasePipelineDataRegistry, metaclass=Singleton):
         Raises
         ------
         RuntimeError
-            If warp / transformation file extension is not ".mat" or ".h5".
+            If warper could not be found in ``extra_input``.
         ValueError
             If ``extra_input`` is None when ``target_data``'s space is native.
 
@@ -329,27 +329,26 @@ class CoordinatesRegistry(BasePipelineDataRegistry, metaclass=Singleton):
                     f"{target_data['space']} space for further computation."
                 )
 
-            # Check for warp file type to use correct tool
-            warp_file_ext = extra_input["Warp"]["path"].suffix
-            if warp_file_ext == ".mat":
+            # Check for warper to use correct tool
+            warper = None
+            for entry in extra_input["Warp"]:
+                if entry["dst"] == "native":
+                    warper = entry["warper"]
+            if warper is None:
+                raise_error(
+                    klass=RuntimeError, msg="Could not find correct warper"
+                )
+            if warper == "fsl":
                 seeds = FSLCoordinatesWarper().warp(
                     seeds=seeds,
                     target_data=target_data,
                     extra_input=extra_input,
                 )
-            elif warp_file_ext == ".h5":
+            elif warper == "ants":
                 seeds = ANTsCoordinatesWarper().warp(
                     seeds=seeds,
                     target_data=target_data,
                     extra_input=extra_input,
-                )
-            else:
-                raise_error(
-                    msg=(
-                        "Unknown warp / transformation file extension: "
-                        f"{warp_file_ext}"
-                    ),
-                    klass=RuntimeError,
                 )
 
         return seeds, labels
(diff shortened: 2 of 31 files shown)


junifer/data/coordinates/_fsl_coordinates_warper.py
```diff
@@ -39,17 +39,31 @@ def warp(
...
 
+        Raises
+        ------
+        RuntimeError
+            If warp file path could not be found in ``extra_input``.
+
         """
         logger.debug("Using FSL for coordinates transformation")
 
+        # Get warp file path
```
fraimondo, line 57: Same as before.

junifer/data/masks/_ants_mask_warper.py
```diff
@@ -83,6 +88,17 @@ def warp(
         if dst == "T1w":
             logger.debug("Using ANTs for mask transformation")
 
+            # Get warp file path
```
fraimondo, line 91: Same as before

junifer/data/masks/_ants_mask_warper.py
```diff
@@ -83,6 +88,17 @@ def warp(
         if dst == "T1w":
```
fraimondo, line 88: why is this T1w here?

junifer/pipeline/pipeline_step_mixin.py
```diff
@@ -196,10 +196,18 @@ def _check_conditional_dependencies(obj) -> None:
                     for dependency in obj._CONDITIONAL_DEPENDENCIES:
                         if dependency["using"] == obj.using:
                             depends_on = dependency["depends_on"]
-                            # Check dependencies
-                            _check_dependencies(depends_on)
-                            # Check external dependencies
-                            _check_ext_dependencies(depends_on)
+                            # Conditional to make `using="auto"` work
```
fraimondo, line 199:
```python
if not isinstance(depends_on, list):
    depends_on = [depends_on]
for entry in depends_on:
    # Check dependencies
    _check_dependencies(entry)
    # Check external dependencies
    _check_ext_dependencies(entry)
 ```

File: junifer/pipeline/update_meta_mixin.py [hunk 5]
```diff
@@ -36,17 +36,35 @@ def update_meta(
...
-        if "dependencies" not in input["meta"]:
-            input["meta"]["dependencies"] = set()
-        # Update step dependencies
-        dependencies = getattr(self, "_DEPENDENCIES", set())
-        if dependencies is not None:
-            if not isinstance(dependencies, (set, list)):
-                dependencies = {dependencies}
-            input["meta"]["dependencies"].update(dependencies)
+        # Conditional for list dtype vals like Warp
+        if isinstance(input, list):
```
fraimondo, line 40: same here, if it's not a list, make it a single element list and avoid duplicated code.

fraimondo (changes requested):
I did not repeat my comment several times, but the logic to find the right warper/transform needs to consider both the src and dst spaces everywhere.

Maybe a util function can sever as such, also checking if more than one transform is available and raising an error in such cases.

The only special case is coordinates, as long as one MNI space is available, it should work.

docs/extending/dependencies.rst
```diff
@@ -78,7 +78,9 @@ that it shows the problem a bit better and how we solve it:
 
     class SpaceWarper(BasePreprocessor):
         # docstring
-        _CONDITIONAL_DEPENDENCIES: ClassVar[List[Dict[str, Union[str, Type]]]] = [
+        _CONDITIONAL_DEPENDENCIES: ClassVar[
+            List[Dict[str, Union[str, Type, List[str], List[Type]]]]
```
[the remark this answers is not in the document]
synchon: That's a fair point. What about having a `junifer.typing` module to take care of this and a lot of other type hinting mess across the codebase?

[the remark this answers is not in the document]
fraimondo: +1

junifer/data/masks/_ants_mask_warper.py
synchon: Because that's provided here: https://github.com/juaml/junifer/blob/2d974e1f1c9c569e5f55f2d7e61460f526d931f0/junifer/data/masks/_masks.py#L536

synchon:
@fraimondo Updated the logic now, would appreciate your feedback.

junifer/data/masks/_ants_mask_warper.py
fraimondo: I meant that it does not make any sense to asume that T1w is native space.

The comment says "native space warping" and then check if `dst = "T1w"`. This would be true if the space of the T1w is native.

junifer/data/masks/_masks.py
fraimondo, line 1: We have an issue with the logic when `mask_type` is `"brain"`, we are returning the `T1w`. We should be returning the `brain` mask.

junifer/preprocess/warping/_fsl_warper.py
```diff
@@ -59,13 +59,28 @@ def preprocess(
...
         logger.debug("Using FSL for space warping")
 
         # Get the min of the voxel sizes from input and use it as the
         # resolution
         resolution = np.min(input["data"].header.get_zooms()[:3])
 
+        # Get warp file path
+        warp_file_path = None
+        for entry in extra_input["Warp"]:
+            if entry["dst"] == "native":
```
fraimondo, line 77: Based on this logic and the `get_native_warper` function I asume that it's not possible for a datagrabber to provide transforms for the same space pair using two different warpers. Is this assumption correct?

fraimondo: If so, can we add the check in the "validate" section of the datagrabber to make it more explicit?

fraimondo (changes requested):
The logic is now good.

There are some comments I made on some issues that are not related to this refactor, but might be worth addressing here.

Like assuming that `dst == "T1w"` means warp to native space.

junifer/data/masks/_masks.py
synchon: When `mask_type="brain"`, `get_template()` fetches the `T1w` template, if we provide `"brain"` to `get_template()`, we'll get the brain mask, which would then be used as `source_img`'s value for `nilearn.image.resample_to_img`, is that the desired outcome?

junifer/preprocess/warping/_fsl_warper.py
synchon: > Based on this logic and the `get_native_warper` function I asume that it's not possible for a datagrabber to provide transforms for the same space pair using two different warpers. Is this assumption correct?

Yes that's why we raise an error now just as we discussed. We can update the logic to take care of that when we actually have the transform files for two different warpers.

synchon: > If so, can we add the check in the "validate" section of the datagrabber to make it more explicit?

Not sure I get you, can you please explain?

junifer/data/masks/_ants_mask_warper.py
synchon: Fair enough.

junifer/data/coordinates/_coordinates.py
```diff
@@ -330,27 +330,26 @@ def get(
                     f"{target_data['space']} space for further computation."
                 )
 
-            # Check for warp file type to use correct tool
-            warp_file_ext = extra_input["Warp"]["path"].suffix
-            if warp_file_ext == ".mat":
+            # Check for warper to use correct tool
+            warper = None
+            for entry in extra_input["Warp"]:
```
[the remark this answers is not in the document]
synchon: Addressed.

junifer/data/masks/_masks.py
fraimondo: Yes, the brain is somehow the union of GM, WM and CSF. The T1w does not make any sense as a "mask", but makes sense as a template.

junifer/preprocess/warping/_fsl_warper.py
fraimondo: I got mixed up. We don't really have a "validator" for post-datagrabbing consistency.

What I wanted to prevent is that someone creates a datagrabber which gives multiple warping options... not for now.

junifer/data/masks/_masks.py
synchon: Alright.

junifer/preprocess/warping/_fsl_warper.py
synchon: Resolving then.

fraimondo (approved):

PR #402, opened 2024-11-25
[ENH]: Allow `workdir` to forward parameters from YAML

synchon:
* [x] tests added/passed

This PR allows `workdir` to be a mapping with `path` and `cleanup` keys in the YAML, which are then forwarded with checks to initialize `WorkDirManager` for debug cases with single element.


fraimondo (approved):
Excellent!

It only allows `cleanup` to be `True` for single-element runs.

PR #409, opened 2024-12-02
[BUG]: Correct file reading for AFNI outputs

synchon:
* [x] tests added/passed

This PR updates the interface for AFNI-based markers and also fixes correct output suffices for AFNI outputs depending on NIfTI headers.

### DIFF
--- a/junifer/markers/falff/falff_base.py
+++ b/junifer/markers/falff/falff_base.py
@@ -146,7 +146,7 @@ class ALFFBase(BaseMarker):
             estimator = JuniferALFF()
         # Compute ALFF + fALFF
         alff, falff, alff_path, falff_path = estimator.compute(  # type: ignore
-            data=input_data["data"],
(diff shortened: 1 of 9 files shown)


junifer/markers/falff/falff_base.py
```diff
@@ -146,7 +146,7 @@ def _compute(
             estimator = JuniferALFF()
         # Compute ALFF + fALFF
         alff, falff, alff_path, falff_path = estimator.compute(  # type: ignore
-            data=input_data["data"],
+            input_path=input_data["path"],
```
fraimondo, line 149: Is there any reason why we move to paths instead of in-memory data?

synchon: Passing the path is easier to reason about inside the method and cheaper to hash for the LRU cache.

fraimondo (approved):

PR #412, opened 2024-12-05
[ENH]: Logging with a color formatter

fraimondo:
Taking advantage of the xterm-colors... this PR adds colors to logging messages.

Screenshot 2024-12-05 at 10 47 17

It also allows to set the datalad logger to a different level than the one from junifer, allowing to debug junifer without the million debugging messages from datalad.


junifer/utils/logging.py
```diff
@@ -297,11 +351,19 @@ def configure_logging(
         #     "%(asctime)s [%(levelname)s] %(message)s "
         #     "(%(filename)s:%(lineno)s)"
         # )
-    formatter = logging.Formatter(fmt=output_format)
+    formatter = ColorFormatter(fmt=output_format)
```
synchon, line 354: What about an option for a `--color` parameter for switching between the formatters? Maybe some remote machine doesn't have xterm support and that might break it?

synchon (changes requested):

junifer/utils/logging.py
```diff
@@ -80,6 +80,59 @@ def __getattr__(self, name: str) -> str:
...
+    Parameters
+    ----------
+    fmt : str
+        The format string for the logging message.
+    datefmt : str, optional
+        The format string for the date.
+
+    """
+
+    BLACK, RED, GREEN, YELLOW, BLUE, MAGENTA, CYAN, WHITE = range(8)
```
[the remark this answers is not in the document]
fraimondo: just copilot being copilot

junifer/utils/logging.py
fraimondo: I will auto-detect color support

synchon (approved):
Needs a newsfragment and it's good to go.

PR #413, opened 2024-12-05
[ENH]: Update `get_template` interface

synchon:
* [x] tests added/passed

This PR replaces the `target_data` parameter of `get_template`, with `target_img`.

### DIFF
--- a/junifer/data/parcellations/_ants_parcellation_warper.py
+++ b/junifer/data/parcellations/_ants_parcellation_warper.py
@@ -131,7 +131,7 @@ class ANTsParcellationWarper:
             # Get template space image
             template_space_img = get_template(
                 space=dst,
-                target_data=target_data,
+                target_img=parcellation_img,
                 extra_input=None,
             )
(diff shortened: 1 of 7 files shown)


junifer/data/parcellations/_ants_parcellation_warper.py
```diff
@@ -131,7 +131,7 @@ def warp(
             # Get template space image
             template_space_img = get_template(
                 space=dst,
-                target_data=target_data,
+                target_img=parcellation_img,
```
fraimondo, line 134: i think this one needs to be in the refactor of the parcellation logic, otherwise it will fail tests here.

fraimondo (changes requested):

junifer/data/parcellations/_ants_parcellation_warper.py
synchon: I updated it for both parcellation and mask to reduce the confusion in the other PRs.

PR #319, opened 2024-12-18
[ENH]: Raise/warn if element is not ran

synchon:

I just did this:

```
  patterns:
    VBM_GM: 
        pattern: derivatives/fmriprep/sub-{subject}/anat/sub-{subject}_space-MNI152NLin2009cAsym_label-GM_probseg.nii.gz
        space: MNI152NLin2009cAsym
  replacements: 
    - subject
```

and then tried to run junifer with `--element sub-0001`. 

The issue (which take me a lot of time to solve) is that `sub-0001` is not in the _elements_ list, since the replacement is without the `sub-` prefix.

Now, there was no issue/warning from the run function. It cloned the dataset, did nothing and dropped the dataset.

We should implement a check in which if some of the elements in the list is not selected by the filter, there's a warning/error.

Exactly here is where we pass through the filter.

https://github.com/juaml/junifer/blob/a9e799caa3433d3d4eaa633d4565771b8eb297f6/junifer/api/functions.py#L170-L178

We should somehow keep a list of "unselected" elements so we can raise the warning.

_No response_

### Anything else to say?

_No response_

### DIFF
--- a/junifer/api/functions.py
+++ b/junifer/api/functions.py
@@ -20,7 +20,13 @@ from ..pipeline import (
 )
 from ..preprocess import BasePreprocessor
 from ..storage import BaseFeatureStorage
-from ..typing import DataGrabberLike, MarkerLike, PreprocessorLike, StorageLike
+from ..typing import (
+    DataGrabberLike,
+    Elements,
+    MarkerLike,
+    PreprocessorLike,
+    StorageLike,
+)
 from ..utils import logger, raise_error, warn_with_log, yaml
 
 
@@ -121,7 +127,7 @@ def run(
     markers: list[dict],
     storage: dict,
     preprocessors: Optional[list[dict]] = None,
-    elements: Optional[list[tuple[str, ...]]] = None,
+    elements: Optional[Elements] = None,
 ) -> None:
     """Run the pipeline on the selected element.
 
@@ -147,7 +153,7 @@ def run(
         List of preprocessors to use. Each preprocessor is a dict with at
         least a key ``kind`` specifying the preprocessor to use. All other keys
         are passed to the preprocessor constructor (default None).
-    elements : list of tuple or None, optional
+    elements : list or None, optional
         Element(s) to process. Will be used to index the DataGrabber
         (default None).
 
@@ -257,7 +263,7 @@ def queue(
     kind: str,
     jobname: str = "junifer_job",
     overwrite: bool = False,
-    elements: Optional[list[tuple[str, ...]]] = None,
+    elements: Optional[Elements] = None,
     **kwargs: Union[str, int, bool, dict, tuple, list],
 ) -> None:
     """Queue a job to be executed later.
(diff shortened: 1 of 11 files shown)


junifer/api/functions.py
```diff
@@ -208,10 +216,22 @@ def run(
     # Fit elements
     with datagrabber_object:
         if elements is not None:
-            for t_element in datagrabber_object.filter(
-                elements  # type: ignore
-            ):
+            # Keep track of valid selectors
```
fraimondo, line 219: This logic won't work in case element is part of a tuple of elements.

The idea is that we can have subject/session/task/etc as the element but only pass the "subject" part an the filter function will return all of the elements for that subject.


I think that the logic should be to check if all items in the `element` array yielded at least one valid element.

synchon: Let me try to make it concrete and maybe it'll be easier to understand:

For this DataGrabber:

```yaml
...
...
...
datagrabber:
  kind: DMCC13Benchmark
  types:
    - BOLD 
...
...
...
```

`junifer list-elements <yaml>` will give you:

```console
...
...
...
sub-f9057kp,ses-wave1pro,Rest,AP,1
sub-f9057kp,ses-wave1pro,Rest,PA,2
sub-f9057kp,ses-wave1rea,Rest,AP,1
sub-f9057kp,ses-wave1rea,Rest,PA,2
```

Now if I do:

```console
junifer run <yaml> --element sub-f9057kp
```
will run all "four" elements of the "subject".

Now, if I instead do:

```console
junifer run <yaml> --element sub-f9057k
```
it'd just "exit" without the changes in this PR, but give you this:
```console
...
...
...
RuntimeError: The following element selectors are invalid:
{'sub-f9057k'}
```
with the current PR. Also, doing this:

```console
junifer run <yaml> --element sub-f9057kp,ses-wave1pr
```
would give you:
```console
...
...
...
RuntimeError: The following element selectors are invalid:
{('sub-f9057kp', 'ses-wave1pr')}
```
Is this what you meant or something else?

fraimondo: That's what I meant. I'm kind of rusty now, but I had the impression that the PR would not account for that logic.

synchon: Resolving it then.

junifer/api/tests/test_functions.py
```diff
@@ -128,18 +147,30 @@ def test_run_single_element_with_preprocessing(
...
         ],
-        elements=[("sub-01",)],
+        elements=["sub-01"],
     )
     # Check files
     files = list(tmp_path.glob("*.sqlite"))
     assert len(files) == 1
 
 
+@pytest.mark.parametrize(
```
fraimondo, line 157: Can you add a test case for the case we discussed?

Ej, sub-01 exists, but ses-02 does not.

fraimondo (changes requested):

fraimondo (approved):

fraimondo:
Once CI is green, LGTM!

PR #418, opened 2025-01-08
[ENH]: Use `junifer-data` datalad dataset for fetching data

synchon:
* [x] tests added/passed

This PR adapts `junifer-data` datalad dataset at https://github.com/juaml/junifer-data to provide data for junifer.


junifer/data/utils.py
```diff
@@ -114,3 +122,96 @@ def get_native_warper(
         )
 
     return possible_warpers[0]
+
+
+def check_dataset() -> dl.Dataset:
```
fraimondo, line 127: This function needs to check that we are also using/cloning the exact junifer data version that we want.

In our last discussion, we said we will have a subdirectory inside the `junifer_data` with the corresponding version to be used with the currently running junifer version.

So somewhere in junifer code, we will have a JUNIFER_DATA_VERSION variable with a string/number/id.

Then we will look if the directory `junifer_data/{JUNIFER_DATA_VERSION}` exists, that its a datalad dataset, it's clean and its tag corresponds to JUNIFER_DATA_VERSION.

This way we can keep junifer data for several installed junifer versions, each one consistently using the data.

junifer/data/utils.py
```diff
@@ -114,3 +122,96 @@ def get_native_warper(
...
+        )
+    else:
+        got_path = Path(got[0]["path"])
+        # Conditional logging based on file fetch
+        status = got[0]["status"]
+        if status == "ok":
+            logger.info(f"Successfully fetched file: {got_path.resolve()}")
+            return got_path
+        elif status == "notneeded":
+            logger.info(f"Found existing file: {got_path.resolve()}")
```
fraimondo, line 211: This should be DEBUG info.

fraimondo (changes requested):

synchon:
Needs to go before #363

PR #363, opened 2025-01-21
[ENH]: Move data downloading/handling to `junifer-data` package.

synchon:

Currently, Junifer is handling the data, downloading from various sources. Ideally, this should be centralized and versiones.

1. Junifer uses the `junifer-data` API.
2. `junifer-data` keeps a datalad dataset with all data objects, including externals. Each commit to this datasets has a tag.
3. `junifer` knows exactly which is the tag that we are dealing with, so each junifer version will always work with the same data files.
4. We have a command `download` or similar that makes junifer download all (or the requested) files to be used later on with the `run` command.

_No response_

### Anything else to say?

_No response_

### DIFF
--- a/junifer/data/utils.py
+++ b/junifer/data/utils.py
@@ -22,7 +22,7 @@ __all__ = [
 
 
 # junifer-data version constant
-JUNIFER_DATA_VERSION = "1.0.0"
+JUNIFER_DATA_VERSION = "1"
 
 
 def closest_resolution(
--- a/junifer/data/utils.py
+++ b/junifer/data/utils.py
@@ -8,19 +8,21 @@ from collections.abc import MutableMapping
 from pathlib import Path
 from typing import Optional, Union
 
-import datalad.api as dl
 import numpy as np
-from datalad.support.exceptions import IncompleteResultsError
 
 from ..utils import config, logger, raise_error
 
 
 __all__ = [
-    "check_dataset",
+    "JUNIFER_DATA_VERSION",
     "closest_resolution",
-    "fetch_file_via_datalad",
+    "get_dataset_path",
     "get_native_warper",
 ]
+
+
+# junifer-data version constant
+JUNIFER_DATA_VERSION = "1.0.0"
 
 
 def closest_resolution(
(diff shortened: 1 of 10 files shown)


synchon:
> > 4. We have a command download or similar that makes junifer download all (or the requested) files to be used later on with the run command.
> 
> This is not yet implemented and would depend on `junifer-data` API. Will implement after the necessary API is available from `junifer-data`.

`junifer-data`-client v1.1.0 now has a `download` command to do this via CLI only.

junifer/data/utils.py
```diff
@@ -5,14 +5,24 @@
...
+__all__ = [
+    "JUNIFER_DATA_VERSION",
+    "closest_resolution",
+    "get_dataset_path",
+    "get_native_warper",
+]
+
+
+# junifer-data version constant
+JUNIFER_DATA_VERSION = "1.0.0"
```
fraimondo, line 25: I would keep this not semver. Just incremental tags. Also easier to handle in case we want to "clean" the directory. Any new modification to the junifer-data package is a new version.

pyproject.toml
```diff
@@ -46,13 +46,13 @@ dependencies = [
...
     "ruamel.yaml>=0.17,<0.19",
     "h5py>=3.10",
-    "httpx[http2]>=0.26.0,<0.28.0",
     "tqdm>=4.66.1,<4.67.0",
     "templateflow>=23.0.0",
     "lapy>=1.0.0,<2.0.0",
     "lazy_loader==0.4",
     "importlib_metadata; python_version<'3.9'",
     "looseversion==1.3.0; python_version>='3.12'",
+    "junifer_data==1.1.0",
```
fraimondo, line 55: Indeed here is confusing. The `junifer_data` package can be updated, but the data can be the same, no? I mean, datalad dataset version is the same.

So in this case, to avoid confusion, i'll have the junifer-data following semver and the dataset version tag as incremental numbers.

fraimondo (changes requested):

junifer/data/utils.py
synchon: No problem with no semver and just int. Not sure about the rationale of easier "clean".

pyproject.toml
synchon: `junifer_data` is the client API which adds and improves the API we use to interact with the `junifer-data` datalad dataset.

- `junifer_data` in `pyproject.toml` refers to: https://github.com/juaml/junifer-data-client
- `JUNIFER_DATA_VERSION` in `junifer/data/utils.py` refers to https://github.com/juaml/junifer-data

fraimondo: perfect

fraimondo (approved):

fraimondo:
CI is not green though

synchon:
> CI is not green though

It's a new problem with gh-action where numpy is not getting built for the docs. Will take care of it after getting the important stuff in.

PR #421, opened 2025-01-23
[ENH]: Add scrubbing support in `fMRIPrepConfoundRemover`

synchon:
* [x] tests added/passed

This PR adds scrubbing support in `fMRIPrepConfoundRemover` by using `std_dvars` from fMRIPrep output.

### DIFF
--- a/junifer/preprocess/confounds/fmriprep_confound_remover.py
+++ b/junifer/preprocess/confounds/fmriprep_confound_remover.py
@@ -489,18 +489,24 @@ class fMRIPrepConfoundRemover(BasePreprocessor):
             and "std_dvars" not in confounds_df.columns
         ):
             raise_error(
-                "Invalid confounds file. Missing std_dvars "
-                "(standardized DVARS) confound. "
-                "Check if this file is really an fMRIPrep confounds file. "
+                msg=(
+                    "Invalid confounds file. Missing std_dvars "
+                    "(standardized DVARS) confound. "
+                    "Check if this file is really an fMRIPrep confounds file. "
+                ),
+                klass=RuntimeError,
             )
         if (
             self.fd_threshold is not None
             and "framewise_displacement" not in confounds_df.columns
         ):
             raise_error(
-                "Invalid confounds file. Missing framewise_displacement "
-                "confound. "
-                "Check if this file is really an fMRIPrep confounds file. "
+                msg=(
+                    "Invalid confounds file. Missing framewise_displacement "
+                    "confound. "
+                    "Check if this file is really an fMRIPrep confounds file. "
+                ),
+                klass=RuntimeError,
             )
         # Use function from nilearn to not reinvent the wheel
         return _load_scrub(
--- a/junifer/preprocess/confounds/fmriprep_confound_remover.py
+++ b/junifer/preprocess/confounds/fmriprep_confound_remover.py
@@ -701,10 +701,16 @@ class fMRIPrepConfoundRemover(BasePreprocessor):
             )
         # Clean image
         logger.info("Cleaning image using nilearn")
+        logger.debug(f"\tstrategy: {self.strategy}")
+        logger.debug(f"\tspike: {self.spike}")
+        logger.debug(f"\tscrub: {self.scrub}")
+        logger.debug(f"\tfd_threshold: {self.fd_threshold}")
+        logger.debug(f"\tstd_dvars_threshold: {self.std_dvars_threshold}")
         logger.debug(f"\tdetrend: {self.detrend}")
         logger.debug(f"\tstandardize: {self.standardize}")
         logger.debug(f"\tlow_pass: {self.low_pass}")
         logger.debug(f"\thigh_pass: {self.high_pass}")
+        logger.debug(f"\tt_r: {self.t_r}")
 
         # Deconfound data
         cleaned_img = nimg.clean_img(
--- a/junifer/preprocess/confounds/fmriprep_confound_remover.py
+++ b/junifer/preprocess/confounds/fmriprep_confound_remover.py
@@ -627,11 +627,15 @@ class fMRIPrepConfoundRemover(BasePreprocessor):
                 }
             )
 
+        signal_clean_kwargs = {}
         # Set up scrubbing mask if needed
         if self.std_dvars_threshold is not None:
-            sample_mask = self._get_scrub_mask(input["confounds"])
-        else:
-            sample_mask = None
+            sample_mask = self._get_scrub_mask(confounds_df)
+            signal_clean_kwargs.update(
+                {
+                    "clean__sample_mask": sample_mask,
+                }
+            )
 
         # Clean image
         logger.info("Cleaning image using nilearn")
@@ -650,7 +654,7 @@ class fMRIPrepConfoundRemover(BasePreprocessor):
             high_pass=self.high_pass,
             t_r=t_r,
             mask_img=mask_img,
-            clean__sample_mask=sample_mask,
+            **signal_clean_kwargs,
         )
         # Fix t_r as nilearn messes it up
         cleaned_img.header["pixdim"][4] = t_r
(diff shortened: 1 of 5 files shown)


junifer/preprocess/confounds/fmriprep_confound_remover.py
```diff
@@ -404,6 +423,14 @@ def _pick_confounds(self, input: dict[str, Any]) -> pd.DataFrame:
             out_df["spike"] = fd
             to_select.append("spike")
 
+        # add binary std_dvars regressor if needed at given threshold
```
fraimondo, line 426: std_dvars should not be added as a regressor to perform scrubbing, but used to generate a mask that is then passed on to `signal.clean`

fraimondo (changes requested):

junifer/preprocess/confounds/fmriprep_confound_remover.py
synchon: So the mask would be temporal which should remove the timepoints from both BOLD image and confounds table and then pass the result to `nilearn.image.clean_img()`?

fraimondo:
Two comments:

1) Why using DVARS? where did you get the reference for this method? I've seen scrubbing also using the FD
2) As it is, it nows also adds DVARS as a regressor.

I think the implementation should come from nilearn: https://nilearn.github.io/stable/modules/generated/nilearn.interfaces.fmriprep.load_confounds.html#nilearn.interfaces.fmriprep.load_confounds

Basically here you can set the `fd_threshold` and `std_dvars_threshold` and it will give you the `sample_mask`

synchon:
> Two comments:
> 
> 1. Why using DVARS? where did you get the reference for this method? I've seen scrubbing also using the FD

`nilearn.image.load_confounds` has a parameter `std_dvars_threshold` for scrubbing, so I went for it as well. We support FD via the `spike` parameter of `fMRIPrepConfoundRemover`.

> 2. As it is, it nows also adds DVARS as a regressor.

It generates a mask for indexing the time dimension and passes it to `nilearn.image.clean` as you suggested in your previous comment.

> I think the implementation should come from nilearn: https://nilearn.github.io/stable/modules/generated/nilearn.interfaces.fmriprep.load_confounds.html#nilearn.interfaces.fmriprep.load_confounds
> 
> Basically here you can set the `fd_threshold` and `std_dvars_threshold` and it will give you the `sample_mask`

We don't use `load_confounds` but select the confounds we want to remove and pass it to `nilearn.image.clean_img`. Also, `load_confounds` relies on the confound file by following the path from the image file. Is that something we go for?

fraimondo:
> > Two comments:
> > 
> > 1. Why using DVARS? where did you get the reference for this method? I've seen scrubbing also using the FD
> 
> `nilearn.image.load_confounds` has a parameter `std_dvars_threshold` for scrubbing, so I went for it as well. We support FD via the `spike` parameter of `fMRIPrepConfoundRemover`.

`spike` is to add as a regressor, which is not scrubbing: https://github.com/juaml/junifer/blob/e392c7bef75d7ba3bc49a44b06b596f576fc4543/junifer/preprocess/confounds/fmriprep_confound_remover.py#L411

> 
> > 2. As it is, it nows also adds DVARS as a regressor.
> 
> It generates a mask for indexing the time dimension and passes it to `nilearn.image.clean` as you suggested in your previous comment.

I see that you use `input["confounds"]` for the scrub mask and not `confounds_df`. We are good as long as dvars or any confounds used for the scrubbing appear in the confounds of `clean_img`

> 
> > I think the implementation should come from nilearn: https://nilearn.github.io/stable/modules/generated/nilearn.interfaces.fmriprep.load_confounds.html#nilearn.interfaces.fmriprep.load_confounds
> > Basically here you can set the `fd_threshold` and `std_dvars_threshold` and it will give you the `sample_mask`
> 
> We don't use `load_confounds` but select the confounds we want to remove and pass it to `nilearn.image.clean_img`. Also, `load_confounds` relies on the confound file by following the path from the image file. Is that something we go for?

We need to somehow "mimic" that behaviour. That's what I meant. The scrubbing mask can be either from dvars or fd.

synchon:
> I see that you use input["confounds"] for the scrub mask and not confounds_df. We are good as long as dvars or any confounds used for the scrubbing appear in the confounds of clean_img

Do the new commits address it?

fraimondo:
> > I see that you use input["confounds"] for the scrub mask and not confounds_df. We are good as long as dvars or any confounds used for the scrubbing appear in the confounds of clean_img
> 
> Do the new commits address it?

We are still missing the `fd_threshold` and the `scrub` parameter. Also check the reference for scrubbing: https://www.sciencedirect.com/science/article/abs/pii/S1053811913009117?via%3Dihub

Coverage is low, meaning that we are missing tests.

synchon:
> We are still missing the fd_threshold and the scrub parameter.

Just so that it's explicit:

- Does fMRIPrepConfoundRemover's `strategy` take a key named `scrub` or does one pass it via the parameters?
- From `load_confounds` docs: `“scrub” regressors for Power et al.[[3]](https://nilearn.github.io/stable/modules/generated/nilearn.interfaces.fmriprep.load_confounds.html#footcite-power2014) scrubbing approach. Associated parameter: scrub, fd_threshold, std_dvars_threshold`. It treats them as "regressors" and when you pass `"scrub"` in the `strategy`, it does not return either `framewise_displacement` or `std_dvars` in the returned confounds dataframe.

> We are good as long as dvars or any confounds used for the scrubbing appear in the confounds of clean_img

I thought the columns should be in the returned dataframe?
-  If you are ok with how the `std_dvars_threshold` is currently implemented for fMRIPrepConfoundRemover, I can add `fd_threshold` as well. I'm not sure how `scrub` is supposed to be implemented?

synchon:
@fraimondo would really appreciate your feedback to the previous comment.

synchon:
I've updated the interface and also improved the logic by replicating nilearn's implementation, would appreciate a review @fraimondo @kaurao

junifer/preprocess/confounds/fmriprep_confound_remover.py
```diff
@@ -578,12 +679,38 @@ def preprocess(
...
             )
+
+        signal_clean_kwargs = {}
+        # Set up scrubbing mask if needed
+        if self.strategy.get("scrubbing", False):
+            motion_outline_regressors = self._get_scrub_regressors(
+                input["confounds"]["data"]
+            )
+            # Add regressors to confounds
+            confounds_df = pd.concat(
```
fraimondo, line 690: Here you add the required variables for scrubbing to the df. Are they used later as regressors? Or they are not considered?

I fear that we might be regressing out this variables too.

junifer/preprocess/confounds/tests/test_fmriprep_confound_remover.py
```diff
@@ -567,3 +567,41 @@ def test_fMRIPrepConfoundRemover_fit_transform_masks() -> None:
...
+
+        # Data should have different shape
+        assert_raises(
+            AssertionError,
+            assert_array_equal,
+            orig_bold.shape,
+            trans_bold.shape,
+        )
+        # and be different
+        assert_raises(
```
fraimondo, line 605: You can actually check the thresholds and which samples should be removed here.

junifer/preprocess/confounds/fmriprep_confound_remover.py
synchon: Motion outlier regressors are generated using the thresholds: https://github.com/nilearn/nilearn/blob/7a8cd0ee7161c4ea5ce4c2df248f0113890e3070/nilearn/interfaces/fmriprep/load_confounds_components.py#L222 and then added to the df: https://github.com/nilearn/nilearn/blob/7a8cd0ee7161c4ea5ce4c2df248f0113890e3070/nilearn/interfaces/fmriprep/load_confounds.py#L484 which then gets processed to generate the sample mask and the added motion outlier regressors are removed to return the proper confounds. This is leveraging nilearn's implementation so no code of ours should affect it.

fraimondo: ok then.

fraimondo (approved):
Once CI is ready, we can merge

PR #424, opened 2025-01-27
FIx CI coverage

fraimondo:
* [x] fix #(issue number)
* [x] tests added/passed

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -52,7 +52,7 @@ deps =
     pytest
     pytest-cov
 commands =
-    pytest --cov={envsitepackagesdir}/junifer --cov-report=xml --cov-report=term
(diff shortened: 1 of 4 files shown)


fraimondo:
Just a fix for the coverage issue.

tox.ini
```diff
@@ -52,7 +52,7 @@ deps =
     pytest
     pytest-cov
 commands =
-    pytest --cov={envsitepackagesdir}/junifer --cov-report=xml --cov-report=term
```
fraimondo, line 55: We either test and evaluate coverage on `{envsitepackagesdir}` or on the _checked out_ code.

My take is that this way we are testing what we ship in the installable package.

synchon: Yeah it stopped reporting coverage out of the blue so as long as we are back, we are good.

synchon (approved):
Needs a fragment file and good to go.

PR #333, opened 2025-03-18
[ENH]: Support "xi" correlation metric for functional connectivity

synchon:

Adapt https://github.com/czbiohub-sf/xicor to be used as a correlation metric for functional connectivity markers.

As an option for nilearn's ConnectivityMeasure.

```shell
https://github.com/czbiohub-sf/xicor
```

### Anything else to say?

_No response_

### DIFF
--- a/docs/changes/newsfragments/333.enh
+++ b/docs/changes/newsfragments/333.enh
@@ -1,0 +1,1 @@
+Add support for Chatterjee's xi correlation in :class:`.JuniferConnectivityMeasure` by `Synchon Mandal`_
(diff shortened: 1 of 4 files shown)


fraimondo:
If the implementation is the same, then off we go!

I think we needed to allow for custom metrics for this to happen.

synchon:
> If the implementation is the same, then off we go!

Implementation is exact and tuned for scipy which is a win.

> I think we needed to allow for custom metrics for this to happen.

The custom object is `JuniferConnectivityMeasure` which is in main for quite some time now. It also enabled us to add Spearman's correlation.

fraimondo (approved):

PR #441, opened 2025-03-24
[MAINT]: Repository maintenance

synchon:
* [x] tests added/passed

This PR does general repository maintenance by updating `.pre-commit-config.yaml`, replacing `black` with `ruff-format` and updating tool configs.


fraimondo (commented):
Doesn't this PR needs to be rebased or merged on top of another? Seems that its adding covereage and adding the coordinates registry.

synchon:
This PR is already rebased on main and it's removing the leftover files from coordinates and masks.

synchon:
@fraimondo should be good for review now, the patch coverage is red but it's not a problem.

synchon:
@fraimondo Can we get this in or do you have some review comments?

fraimondo (approved):

Issue #296, opened 2024-01-26
[ENH]: Validate boolean parameters passed from yaml file


tomug:
I erroneously passed a list to the DataladHCP1200 data grabber trying to set the `ica_fix` parameter to `false` but provided it in list notation. The data grabber ended up getting the FIX-cleaned data, because the list was resolved to true.

Suggested solve: validate the parameters so if a boolean is expected and a list is given one gets an error.

PR #364, opened 2025-11-13
[ENH]: Adopt Pydantic for schema validation

synchon:

We want to be able to validate the yaml so we avoid issues like #296 

Use pydantic and validate the yaml.

_No response_

### Anything else to say?

_No response_


docs/extending/parcellations.rst
```diff
@@ -124,7 +124,7 @@ parcellation when registering it. For example, we can add a
   markers:
     - name: CustomParcellation_mean
       kind: ParcelAggregation
-      parcellation: my_custom_parcellation
+      parcellation: [<my_custom_parcellation>]
```
fraimondo: Same here: str or list[str]

docs/using/codeless.rst
```diff
@@ -171,11 +171,11 @@ For the ``Oasis VBM Testing dataset`` example, we want to compute the mean
   markers:
     - name: Schaefer100x7_mean
       kind: ParcelAggregation
-      parcellation: Schaefer100x7
+      parcellation: [Schaefer100x7]
```
fraimondo: I will stop commenting on this

junifer/datagrabber/aomic/id1000.py
```diff
@@ -24,212 +28,219 @@ class DataladAOMICID1000(PatternDataladDataGrabber):
...
-        (default None).
-    space : {"native", "MNI152NLin2009cAsym"}, optional
-        The space to use for the data (default "MNI152NLin2009cAsym").
-
-    Raises
-    ------
-    ValueError
-        If invalid value is passed for:
-         * ``space``
+    types : list of {``DataType.BOLD``, ``DataType.T1w``, \
```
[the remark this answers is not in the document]
fraimondo: Same for all the datagrabbers, markers, steps, etc where it can be one or many.

junifer/datagrabber/base.py
```diff
@@ -7,57 +7,73 @@
...
-        self.types = types
-
-        # Convert str to Path
-        if not isinstance(datadir, Path):
-            datadir = Path(datadir)
-        self._datadir = datadir
+    model_config = ConfigDict(use_enum_values=True)
 
+    types: list[DataType] = Field(frozen=True)
+    datadir: Path
```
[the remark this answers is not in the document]
fraimondo: I think that is why it's a property.

junifer/datagrabber/hcp1200/datalad_hcp1200.py
```diff
@@ -23,50 +25,37 @@ class DataladHCP1200(DataladDataGrabber, HCP1200):
...
-        HCP phase encoding directions. If None, both will be used
-        (default None).
+    types : list of {``DataType.BOLD``, ``DataType.T1w``, ``DataType.Warp``}, \
+            optional
+        The data type(s) to grab.
+    datadir : pathlib.Path, optional
+        That path where the datalad dataset will be cloned.
+        If not specified, the datalad dataset will be cloned into a temporary
+        directory.
+    tasks : list of :enum:`.HCP1200Task`, optional
```
fraimondo: will this expand? Otherwise I prefer the previous docstring as it tells you exactly what your options are.

junifer/datagrabber/pattern_datalad.py
```diff
@@ -23,122 +25,24 @@ class PatternDataladDataGrabber(DataladDataGrabber, PatternDataGrabber):
...
+        The datagrabber patterns. Check :class:`DataTypeSchema` for the schema.
+    replacements : list of str
+        All possible replacements in ``patterns.<data_type>.pattern``.
+    rootdir : pathlib.Path, optional
         The path within the datalad dataset to the root directory
-        (default ".").
-    uri : str or None, optional
-        URI of the datalad sibling (default None).
+        (default Path(".")).
+    confounds_format : :enum:`.ConfoundsFormat` or None, optional
```
fraimondo: Having an enum for the confounds format will also restrict someone defining their own format and using their own preprocessor for that format.

fraimondo (changes requested):
Overall looks promising. Though there are some changes that go against the easy-to-use QA:

* Many parameters were `str or list of str`. Now they are lists. This adds the unnecessary burden to add a list in the YAMLs where the most common use case is to use a single element.

* some parameters were `str or Path` and now they have to be Paths. This also makes it more complicated for users to define code. This can easily be validated later on and not with pydantic. Same for `HttpUrl`

* I think there's an issue with the `datadir` which now will be part of the meta due to missing it's `_`. Also, there was a reason why it was not called `fulldir` in the datababber, but `datadir`, on purpose.

* Some enums might add validation to pydantic, but will restraint the extensibility of junifer. E.g. DataType, ConfoundFormat, etc... These should be dynamic, as I might want to add some of them in an extension, without modifying junifer code.


Since there are 197 files changed, and many changes happen because of the first item. I will review more later with the next changes.

junifer/datagrabber/datalad_base.py
```diff
@@ -106,43 +143,27 @@ def __init__(
...
         """Destructor."""
-        if hasattr(self, "_tmpdir"):
-            self._rmtmpdir()
-
-    def _rmtmpdir(self) -> None:
-        """Remove temporary directory if it exists."""
-        if self._tmpdir.exists():
-            logger.debug("Removing temporary directory")
-            WorkDirManager().delete_tempdir(self._tmpdir)
+        if self.datadir.stem.startswith("datalad"):
```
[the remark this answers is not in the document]
synchon: Hmm that's a fair argument, I can do something like:
```python
datadir = WorkDirManager().get_tempdir(
        prefix="datalad", suffix="juniferauto"
)
```
and then check with:
```python
if self.datadir.stem.startswith(
    "datalad"
) and self.datadir.stem.endswith("juniferauto"):
```

synchon:
> * Many parameters were `str or list of str`. Now they are lists. This adds the unnecessary burden to add a list in the YAMLs where the most common use case is to use a single element.

All of them should be reverted now, let me know if you find any stray ones.

> * some parameters were `str or Path` and now they have to be Paths. This also makes it more complicated for users to define code.

Not at all, users can keep passing `str` and they'll be converted to `Path` by pydantic.

> This can easily be validated later on and not with pydantic. Same for `HttpUrl`

The introduction of `pydantic` is primarily to not hand-roll these validations anymore and users should only need to add logical validations, but anyway these can be easily added via pydantic. `HttpUrl` is now `AnyUrl` and is same as the `Path` argument above.

> * I think there's an issue with the `datadir` which now will be part of the meta due to missing it's `_`.

Has been addressed in the specific comment.

> Also, there was a reason why it was not called `fulldir` in the datababber, but `datadir`, on purpose.

The reason was to keep it in sync with non-datalad based datagrabbers and is already taken care of internally via introduction of `_repodir`. The UX should be exactly as it is now, let me know if you find any difference while using.

> * Some enums might add validation to pydantic, but will restraint the extensibility of junifer. E.g. DataType, ConfoundFormat, etc... These should be dynamic, as I might want to add some of them in an extension, without modifying junifer code.

Has been addressed in specific comments.

junifer/datagrabber/datalad_base.py
```diff
@@ -179,10 +202,11 @@ def _get_dataset_id_remote(self) -> tuple[str, bool]:
                     is_dirty = False
             else:
                 logger.debug("Skipping dataset ID check")
+                # Should be already set to the dataset
                 remote_id = self._dataset.id
                 is_dirty = False
             logger.debug(
-                f"Remote dataset is {'' if is_dirty else 'not'} dirty"
+                f"Remote dataset is{'' if is_dirty else ' not '}dirty"
```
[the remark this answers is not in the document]
fraimondo:
```
[junifer] pool-49-54 ➜ ~  ipython
Python 3.11.5 | packaged by conda-forge | (main, Aug 27 2023, 03:33:12) [Clang 15.0.7 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.16.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: is_dirty = True

In [2]: print(f"Remote dataset is{'' if is_dirty else ' not '}dirty")
Remote dataset isdirty

In [3]: is_dirty = False

In [4]: print(f"Remote dataset is{'' if is_dirty else ' not '}dirty")
Remote dataset is not dirty
```

junifer/datagrabber/pattern_datalad.py
synchon: You'd do as shown here: https://juaml.github.io/junifer/pr-preview/pr-364/extending/data_types.html

fraimondo: Can you implement the possibility to extend confound formats?

synchon:
@fraimondo I've pushed the further changes, feel free to review.

docs/extending/preprocessor.rst
```diff
@@ -165,15 +158,9 @@ decorator and our final code should look like this:
 
         _DEPENDENCIES = {"nilearn"}
 
-        _VALID_DATA_TYPES: ClassVar[Sequence[str]] = ["T1w", "T2w", "BOLD"]
+        _VALID_DATA_TYPES: ClassVar[Sequence[DataType]] = ["T1w", "T2w", "BOLD"]
```
fraimondo, line 161: Since they are datatypes, shouldn't they be DataType.T1w, etc? Or it is ok to keep them as strings?

docs/extending/preprocessor.rst
```diff
@@ -204,9 +191,12 @@ Template for a custom Preprocessor
         # TODO: add the inputs
         _VALID_DATA_TYPES = []
```
fraimondo: We can already leave the typing in the template here

junifer/cli/tests/data/gmd_mean_htcondor.yaml
```diff
@@ -17,4 +17,5 @@ queue:
   env:
     kind: conda
     name: junifer
+    shell: bash
```
fraimondo, line 20: We don't have bash as default here?

junifer/datagrabber/aomic/id1000.py
```diff
@@ -7,228 +7,239 @@
...
-            types=types,
-            datadir=datadir,
-            uri=uri,
-            patterns=patterns,
-            replacements=replacements,
-            confounds_format="fmriprep",
+    def validate_datagrabber_params(self) -> None:
+        """Run extra logical validation for datagrabber."""
+        # Descriptor for space in `anat`
+        sp_anat_desc = (
```
fraimondo, line 208: This will happen a alot, might be worth having a "util" function to "update patterns".

junifer/markers/parcel_aggregation.py
```diff
@@ -65,67 +90,61 @@ class ParcelAggregation(BaseMarker):
...
+    parcellation: Annotated[str | list[str], BeforeValidator(ensure_list)]
+    method: str = "mean"
+    method_params: dict[str, Any] | None = None
+    time_method: str | None = None
+    time_method_params: dict[str, Any] | None = None
+    masks: Annotated[
+        dict | str | list[dict | str] | None,
+        BeforeValidator(ensure_list_or_none),
+    ] = None
+    on: Annotated[
```
fraimondo, line 131: Do we still need the `on` here? It's inherited from `BaseMarker`

fraimondo: There are many ocasions in which the `on` is repeated on the inherited marker.

fraimondo (commented):
Last questions

docs/extending/preprocessor.rst
synchon: They get automatically converted.

docs/extending/preprocessor.rst
synchon: What do you mean?

junifer/cli/tests/data/gmd_mean_htcondor.yaml
synchon: Yes the default is bash, this is to be explicit.

junifer/markers/parcel_aggregation.py
synchon: Yes we need it to constrain the correct data types else any DataType will be valid.

junifer/datagrabber/aomic/id1000.py
synchon: I agree it has value and it's a dict so it should be straightforward to make one. I wouldn't want to do it in this PR though, might be good to tackle in another one.

docs/extending/preprocessor.rst
fraimondo: This is supposed to be a template to create your own preprocessor:

```
_VALID_DATA_TYPES: ClassVar[Sequence[DataType]] = []
```

junifer/cli/tests/data/gmd_mean_htcondor.yaml
fraimondo: ok

junifer/datagrabber/aomic/id1000.py
fraimondo: Before it was part of the "init" of each class, but now it has became something more "systematic" in the validation, which also makes the term validation a bit too loose. If you dont' want it in the PR, make an issue.

junifer/markers/parcel_aggregation.py
fraimondo: thought so.

fraimondo (approved):
One checks pass, good to go.

junifer/datagrabber/aomic/id1000.py
synchon: Opened here: https://github.com/juaml/junifer/issues/497

docs/extending/preprocessor.rst
synchon: Addressed in latest commit.

PR #498, opened 2026-05-29
[ENH]: Introduce `junifer.api.generate_yaml`

synchon:
* [ ] tests added/passed

This PR adds `generate_yaml` under `api` to generate feature YAML from metadata. Its primary use-case is in `julio`'s feature addition to registry.


fraimondo:
I have the impression that this way (`dump_exclude`) adds a lot of maintenance work: if something changes in the superclass (new internal variable), we need to go and update every subclass, including non-junifer ones.

Can't we just look at what _fields_ are defined in the class (not the superclass) and just pass those ones to the constructor?

synchon:
> I have the impression that this way (`dump_exclude`) adds a lot of maintenance work: if something changes in the superclass (new internal variable), we need to go and update every subclass, including non-junifer ones.

That's a fair argument and I see your point.
 
> Can't we just look at what _fields_ are defined in the class (not the superclass) and just pass those ones to the constructor?

Not with how I understand the thing works. A model's fields consist of its own fields and superclass' fields (if it has one). So apart from defining what to exclude (or include), I don't see other way. I'll push some updates to make it better.

synchon:
@fraimondo I've updated the datagrabber dumping logic as discussed. Kindly review #499 before this.

junifer/configs/juseless/datagrabbers/ucla.py
```diff
@@ -143,6 +143,11 @@ class JuselessUCLA(PatternDataGrabber):
     replacements: list[str] = ["subject", "task"]  # noqa: RUF012
     confounds_format: ConfoundsFormat = ConfoundsFormat.FMRIPrep
 
+    @classmethod
+    def dump_fields(cls) -> list[str]:
+        """Fields to include when dumping model."""
+        return [*super().dump_fields(), "tasks"]
```
fraimondo: This should only be `types` and `tasks`. The rest is hard-coded in the parameters.

junifer/datagrabber/aomic/id1000.py
fraimondo, line 66: Can this be changed in the constructor?

junifer/datagrabber/hcp1200/datalad_hcp1200.py
```diff
@@ -61,6 +61,11 @@ class DataladHCP1200(DataladDataGrabber, HCP1200):
     ]
     rootdir: Path = Path("HCP1200")
 
+    @classmethod
+    def dump_fields(cls) -> list[str]:
+        """Fields to include when dumping model."""
+        return ["types", "tasks", "phase_encodings", "ica_fix"]
```
fraimondo, line 67: This can also be `super(HCP1200) - datadir`. Thus any change in the super will also be accounted for here.

junifer/datagrabber/datalad_base.py
```diff
@@ -176,6 +176,11 @@ def __del__(self) -> None:
         ) and self.datadir.stem.endswith("juniferauto"):
             _remove_datadir(self.datadir)
 
+    @classmethod
+    def dump_fields(cls) -> list[str]:
+        """Fields to include when dumping model."""
+        return ["types", "uri", "rootdir"]
```
fraimondo, line 182: this is where it becomes a bit tricky.

If I used a datadir (not temp) and the dataset is dirty, the YAML should account for that? or not?

Maybe we should include some comments in the generated YAMLs indicating stuff like this:

eg. if we have a "dirty" dataset, then add a comment that while the yaml will reproduce the results, the original dataset was "dirty" and so there is no guarantee that the same results will be obtained as there is no strict data provenance.

junifer/datagrabber/pattern_datalad.py
```diff
@@ -61,3 +61,16 @@ def validate_datagrabber_params(self) -> None:
         logger.debug("Initializing PatternDataladDataGrabber")
         for key, val in self.__pydantic_extra__.items():
             logger.debug(f"\t{key} = {val}")
+
+    @classmethod
+    def dump_fields(cls) -> list[str]:
+        """Fields to include when dumping model."""
+        return [
```
fraimondo, line 68: I thinks this should be both "super" fields.

junifer/datagrabber/aomic/id1000.py
synchon: Yes

junifer/datagrabber/hcp1200/datalad_hcp1200.py
synchon: In that case, the MRO for this class will get `dump_fields` from `DataladDataGrabber` which will be incorrect.

junifer/datagrabber/pattern_datalad.py
synchon: It is "both" of them.

junifer/datagrabber/datalad_base.py
synchon: We can add a general comment. Making it conditional would be quite tricky.

fraimondo: I would like that the generated YAML is commented.

junifer/datagrabber/pattern_datalad.py
fraimondo: I meant instead of manually placing the fields, to all super. But I understand that might be bothersome.

fraimondo:
Do we know how this function works when we have external imports (`with` statements in the YAML)?

junifer/datagrabber/pattern_datalad.py
synchon: The MRO would stop at the first `dump_fields` which would give a partial list.

synchon:
> Do we know how this function works when we have external imports (`with` statements in the YAML)?

It copies exactly what is passed. For julio, we copy exactly what is stored in h5.

fraimondo:
> > Do we know how this function works when we have external imports (`with` statements in the YAML)?
> 
> It copies exactly what is passed. For julio, we copy exactly what is stored in h5.

In order to create the yaml, we parse the metadata and instantiate the object. I'm not sure this will be possible if the definition of a datagrabber/marker is in an external package which is part of the `with` statement.

synchon:
> > > Do we know how this function works when we have external imports (`with` statements in the YAML)?
> > 
> > 
> > It copies exactly what is passed. For julio, we copy exactly what is stored in h5.
> 
> In order to create the yaml, we parse the metadata and instantiate the object. I'm not sure this will be possible if the definition of a datagrabber/marker is in an external package which is part of the `with` statement.

Let me lay it down w.r.t. julio:

- junifer YAML gets parsed by `junifer.api.parse_yaml` which loads the external modules and registers componenets as needed.
- If `with` is present in the junifer YAML, julio adds it directly to the generated YAML.
- As the external components are already registered, `junifer.api.generate_yaml` can load them from component registry and dump as needed.

For using `junifer.api.generate_yaml` without julio, one would need to already register the components beforehand. Now, as it's part of `junifer.api`, we can presume that the user will do it like so.

Does that solve your concern?

fraimondo:
> > > > Do we know how this function works when we have external imports (`with` statements in the YAML)?
> > > 
> > > 
> > > It copies exactly what is passed. For julio, we copy exactly what is stored in h5.
> > 
> > 
> > In order to create the yaml, we parse the metadata and instantiate the object. I'm not sure this will be possible if the definition of a datagrabber/marker is in an external package which is part of the `with` statement.
> 
> Let me lay it down w.r.t. julio:
> 
>     * junifer YAML gets parsed by `junifer.api.parse_yaml` which loads the external modules and registers componenets as needed.
> 
>     * If `with` is present in the junifer YAML, julio adds it directly to the generated YAML.
> 
>     * As the external components are already registered, `junifer.api.generate_yaml` can load them from component registry and dump as needed.
> 
> 
> For using `junifer.api.generate_yaml` without julio, one would need to already register the components beforehand. Now, as it's part of `junifer.api`, we can presume that the user will do it like so.
> 
> Does that solve your concern?

`junifer.api.generate_yaml` has one parameter that is a `meta` dict. This meta dict can be extracted from and HDF5 file, without parsing any yaml. We need to be able to support generating a YAML from any meta dict, no only after parsing the respective yaml.

Even if this means adding all variables int he meta dict and a huge comment stating that "since this is an external datagrabber/marker/etc not part of junifer core, some of this variables might not be needed or should definitely be removed."

synchon:
> > > > > Do we know how this function works when we have external imports (`with` statements in the YAML)?
> > > > 
> > > > 
> > > > It copies exactly what is passed. For julio, we copy exactly what is stored in h5.
> > > 
> > > 
> > > In order to create the yaml, we parse the metadata and instantiate the object. I'm not sure this will be possible if the definition of a datagrabber/marker is in an external package which is part of the `with` statement.
> > 
> > 
> > Let me lay it down w.r.t. julio:
> > ```
> > * junifer YAML gets parsed by `junifer.api.parse_yaml` which loads the external modules and registers componenets as needed.
> > 
> > * If `with` is present in the junifer YAML, julio adds it directly to the generated YAML.
> > 
> > * As the external components are already registered, `junifer.api.generate_yaml` can load them from component registry and dump as needed.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > For using `junifer.api.generate_yaml` without julio, one would need to already register the components beforehand. Now, as it's part of `junifer.api`, we can presume that the user will do it like so.
> > Does that solve your concern?
> 
> `junifer.api.generate_yaml` has one parameter that is a `meta` dict. This meta dict can be extracted from and HDF5 file, without parsing any yaml. We need to be able to support generating a YAML from any meta dict, no only after parsing the respective yaml.
> 
> Even if this means adding all variables int he meta dict and a huge comment stating that "since this is an external datagrabber/marker/etc not part of junifer core, some of this variables might not be needed or should definitely be removed."

I don't follow. What happens if one adds `"with"` key to the "meta" dictionary passed without loading the external components?

fraimondo:
> > > > > > Do we know how this function works when we have external imports (`with` statements in the YAML)?
> > > > > 
> > > > > 
> > > > > It copies exactly what is passed. For julio, we copy exactly what is stored in h5.
> > > > 
> > > > 
> > > > In order to create the yaml, we parse the metadata and instantiate the object. I'm not sure this will be possible if the definition of a datagrabber/marker is in an external package which is part of the `with` statement.
> > > 
> > > 
> > > Let me lay it down w.r.t. julio:
> > > ```
> > > * junifer YAML gets parsed by `junifer.api.parse_yaml` which loads the external modules and registers componenets as needed.
> > > 
> > > * If `with` is present in the junifer YAML, julio adds it directly to the generated YAML.
> > > 
> > > * As the external components are already registered, `junifer.api.generate_yaml` can load them from component registry and dump as needed.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > For using `junifer.api.generate_yaml` without julio, one would need to already register the components beforehand. Now, as it's part of `junifer.api`, we can presume that the user will do it like so.
> > > Does that solve your concern?
> > 
> > 
> > `junifer.api.generate_yaml` has one parameter that is a `meta` dict. This meta dict can be extracted from and HDF5 file, without parsing any yaml. We need to be able to support generating a YAML from any meta dict, no only after parsing the respective yaml.
> > Even if this means adding all variables int he meta dict and a huge comment stating that "since this is an external datagrabber/marker/etc not part of junifer core, some of this variables might not be needed or should definitely be removed."
> 
> I don't follow. What happens if one adds `"with"` key to the "meta" dictionary passed without loading the external components?

Let's asume you open and HDF5 file, you load the meta and then you pass it to `generate_yaml`. At no point there was a `parse_yaml` call, so all external modules will not be loaded. This code will fail as it needs to instantiate the elements in the meta to generate the yaml.

This use case should be considered. In the case that the object can't be instantiated, the fields should be extracted from the meta dict and a comment in the yaml should be added.

synchon:
> > > > > > > Do we know how this function works when we have external imports (`with` statements in the YAML)?
> > > > > > 
> > > > > > 
> > > > > > It copies exactly what is passed. For julio, we copy exactly what is stored in h5.
> > > > > 
> > > > > 
> > > > > In order to create the yaml, we parse the metadata and instantiate the object. I'm not sure this will be possible if the definition of a datagrabber/marker is in an external package which is part of the `with` statement.
> > > > 
> > > > 
> > > > Let me lay it down w.r.t. julio:
> > > > ```
> > > > * junifer YAML gets parsed by `junifer.api.parse_yaml` which loads the external modules and registers componenets as needed.
> > > > 
> > > > * If `with` is present in the junifer YAML, julio adds it directly to the generated YAML.
> > > > 
> > > > * As the external components are already registered, `junifer.api.generate_yaml` can load them from component registry and dump as needed.
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > For using `junifer.api.generate_yaml` without julio, one would need to already register the components beforehand. Now, as it's part of `junifer.api`, we can presume that the user will do it like so.
> > > > Does that solve your concern?
> > > 
> > > 
> > > `junifer.api.generate_yaml` has one parameter that is a `meta` dict. This meta dict can be extracted from and HDF5 file, without parsing any yaml. We need to be able to support generating a YAML from any meta dict, no only after parsing the respective yaml.
> > > Even if this means adding all variables int he meta dict and a huge comment stating that "since this is an external datagrabber/marker/etc not part of junifer core, some of this variables might not be needed or should definitely be removed."
> > 
> > 
> > I don't follow. What happens if one adds `"with"` key to the "meta" dictionary passed without loading the external components?
> 
> Let's asume you open and HDF5 file, you load the meta and then you pass it to `generate_yaml`. At no point there was a `parse_yaml` call, so all external modules will not be loaded. This code will fail as it needs to instantiate the elements in the meta to generate the yaml.
> 
> This use case should be considered. In the case that the object can't be instantiated, the fields should be extracted from the meta dict and a comment in the yaml should be added.

Do the latest commits address your concern?

junifer/api/functions.py
```diff
@@ -600,3 +608,140 @@ def parse_yaml(filepath: str | Path) -> dict:  # noqa: C901
...
+    post = "\nIssues:\n"
+    issue = (
+        "  - `{0}` is not a built-in component and thus could not be "
+        "reproduced. Either replace (or remove) the component or fill it in "
+        "manually.\n"
+    )
+    # Set datagrabber
+    meta_dg = meta["datagrabber"].copy()
+    a = meta_dg.pop("class")
+    try:
```
fraimondo: I would explicity check for the datagrabber being in the registry that relying on a ValueError.

Could be that because of versions mismatchs, some parameters are renamed and then we do have errors but because of other reasons.

junifer/api/functions.py
```diff
@@ -600,3 +608,140 @@ def parse_yaml(filepath: str | Path) -> dict:  # noqa: C901
...
+        post += f"- datagrabber:\n{issue.format(a)}"
+    # Set preprocessor(s)
+    if "preprocess" in meta:
+        y["preprocess"] = []
+        meta_p = meta["preprocess"].copy()
+        if not isinstance(meta_p, list):
+            meta_p = [meta_p]
+        for mp in meta_p:
+            b = mp.pop("class")
+            try:
```
fraimondo: Same here, explicit check

junifer/api/functions.py
```diff
@@ -600,3 +608,140 @@ def parse_yaml(filepath: str | Path) -> dict:  # noqa: C901
...
+                if "- preprocess:\n" in post:
+                    post += f"{issue.format(b)}"
+                else:
+                    post += f"- preprocess:\n{issue.format(b)}"
+    # Set marker
+    meta_m = meta["marker"].copy()
+    c = meta_m.pop("class")
+    y["markers"] = []
+    try:
+        m = PipelineComponentRegistry().get_class(step="marker", name=c)
```
fraimondo, line 727: Same here

junifer/api/functions.py
```diff
@@ -600,3 +608,140 @@ def parse_yaml(filepath: str | Path) -> dict:  # noqa: C901
...
+    )
+    if "dependencies" in meta:
+        for k, v in meta["dependencies"].items():
+            pre += f"{k}=={v}\n"
+    const = (
+        "\nNotes:\n"
+        "- `datadir` is ignored and not reproduced. "
+        "If `datadir` used was not a temporary directory, you will have to "
+        "manually edit this YAML.\n"
+        "- In case the dataset was 'dirty', there is no guarantee that the "
```
fraimondo: This note should only appear if the dataset was dirty (the meta said so)

fraimondo (changes requested):

junifer/api/functions.py
synchon: The check is updated to be precise now. Also, open to go the non-idiomatic route as well.

junifer/api/functions.py
synchon: Updated now.

junifer/api/functions.py
fraimondo: We still rely on a `ValueError`. It should be something like

if component is registered:
     Instantiate and dump
else:
  - add ALL variables in the meta to the yaml
  - Add the note: "<COMPONENT> is not a built-in component and thus could not be properly regenerated. Some of these entries in the YAML section might be redundant and not needed. Please check the documentation/implementation of this specific datagrabber and remove the unnecesary entries."

fraimondo (changes requested):

synchon:
> We still rely on a ValueError. 

`ValueError` is only raised if the component is not registered. `try...except` is the "idiomatic" way to do it. I understand if that doesn't work and it needs to be superfluous.

fraimondo:
> > We still rely on a ValueError.
> 
> `ValueError` is only raised if the component is not registered. `try...except` is the "idiomatic" way to do it. I understand if that doesn't work and it needs to be superfluous.

In that case, if any part of the instantiation raises a ValueError (like would happen if a parameter changes options, or using an old junifer version), then we go to:

* add ALL variables in the meta to the yaml
* Add the note: " is not a built-in component or it failed to instantiate and thus could not be properly regenerated. Some of these entries in the YAML section might be redundant and not needed. Please check the documentation/implementation of this specific datagrabber and remove the unnecesary

synchon:
> > > We still rely on a ValueError.
> > 
> > 
> > `ValueError` is only raised if the component is not registered. `try...except` is the "idiomatic" way to do it. I understand if that doesn't work and it needs to be superfluous.
> 
> In that case, if any part of the instantiation raises a ValueError (like would happen if a parameter changes options, or using an old junifer version), then we go to:
> 
> * add ALL variables in the meta to the yaml
> * Add the note: " is not a built-in component or it failed to instantiate and thus could not be properly regenerated. Some of these entries in the YAML section might be redundant and not needed. Please check the documentation/implementation of this specific datagrabber and remove the unnecesary

`model_construct` does not raise an exception so there will be no error during the model construction.

fraimondo:
Can we validate? I'm worried about using different junifer versions than the one that generated the meta. Or we either go full strict and not allow any mismatch (which will create a problem with julio later on), or we validate the model. Otherwise, variables that do not match will be "ignored" and not "dumped", which might yield a different yaml than expected.

I prefer to have a YAML with a note saying "check your datagrabber/marker/preprocessor due to possible changes in the API" than one without any message that actually works differently than expected.

synchon:
> Can we validate? I'm worried about using different junifer versions than the one that generated the meta. Or we either go full strict and not allow any mismatch (which will create a problem with julio later on), or we validate the model. Otherwise, variables that do not match will be "ignored" and not "dumped", which might yield a different yaml than expected.
> 
> I prefer to have a YAML with a note saying "check your datagrabber/marker/preprocessor due to possible changes in the API" than one without any message that actually works differently than expected.

`model_construct` is now replaced with `model_validate` which will fail for most due to the nature of the metadata and the models. Necessary comments will be added on generation.

fraimondo:
> > Can we validate? I'm worried about using different junifer versions than the one that generated the meta. Or we either go full strict and not allow any mismatch (which will create a problem with julio later on), or we validate the model. Otherwise, variables that do not match will be "ignored" and not "dumped", which might yield a different yaml than expected.
> > I prefer to have a YAML with a note saying "check your datagrabber/marker/preprocessor due to possible changes in the API" than one without any message that actually works differently than expected.
> 
> `model_construct` is now replaced with `model_validate` which will fail for most due to the nature of the metadata and the models. Necessary comments will be added on generation.

I still don't understand why it will fail for "most". As long as you choose the `dump_fields` and pass it to the constructor, this should recreate the same object.

It should fail in case of:
- External components
- Components that changed API and the meta is from a non-compatible version.

All the rest should not fail. Otherwise we are dumping the wrong variables.

synchon:
> > > Can we validate? I'm worried about using different junifer versions than the one that generated the meta. Or we either go full strict and not allow any mismatch (which will create a problem with julio later on), or we validate the model. Otherwise, variables that do not match will be "ignored" and not "dumped", which might yield a different yaml than expected.
> > > I prefer to have a YAML with a note saying "check your datagrabber/marker/preprocessor due to possible changes in the API" than one without any message that actually works differently than expected.
> > 
> > 
> > `model_construct` is now replaced with `model_validate` which will fail for most due to the nature of the metadata and the models. Necessary comments will be added on generation.
> 
> I still don't understand why it will fail for "most". As long as you choose the `dump_fields` and pass it to the constructor, this should recreate the same object.
> 
> It should fail in case of:
> 
> * External components
> * Components that changed API and the meta is from a non-compatible version.
> 
> All the rest should not fail. Otherwise we are dumping the wrong variables.

Ignore my previous reply's "fail" part, it works as intended.

junifer/api/tests/test_functions.py
```diff
@@ -1025,3 +1033,250 @@ def test_parse_yaml_queue_venv_relative(tmp_path: Path) -> None:
...
+                "class": "ExternalMarker",
+                "on": ["BOLD"],
+                "name": "external",
+            },
+            "_element_keys": ["subject", "task"],
+            "name": "BOLD_external",
+        },
+    ],
+)
+def test_generate_yaml(m: dict) -> None:
```
fraimondo: So this tests that the actual function works. Can we test for correctness?

synchon: The latest commit should check for basic correctness.

